PR #4, opened 2019-11-20
Manually add the autorelease badge

carterkozak:


carterkozak:
Will let the robots take care of this

PR #8, opened 2019-11-21
document options in README

ferozco:


carterkozak (approved):

README.md
```diff
@@ -92,6 +92,16 @@ MyServiceMetrics metrics = MyServiceMetrics.of(server.taggedMetrics());
 Meter creationFailures = metrics.failures("create");
 ```
 
+### Options
+Metric definitions can also include options that do not change the overall declaration, but may affect the way it is 
+handled in a particular context.
+```yml
+options:
+  javaPackage: 'com.palantir.my.package' # Specifies under which package Java classes should be generated
+  javaVisibility: packagePrivate # Specifies visibility of generated Utility class. Defaults to public
```
carterkozak, line 101: I wonder if it makes sense to take advantage of conjure enum unknown values for these in order to document supported options in the schema. Not something for today though

carterkozak (approved):

PR #15, opened 2019-11-25
Expose standalone markdown plugin

ferozco:
## Before this PR
We would only generate a markdown summary of the metrics in the current project if it produced an sls service distribution.

## After this PR
==COMMIT_MSG==
Expose `com.palantir.metric-schema-markdown` plugin which allows any project to opt into metric-schema markdown generation.
==COMMIT_MSG==

## Possible downsides?
Increased API surface area


carterkozak:
I think we need a `./META-INF/gradle-plugins/com.palantir.metric-schema-markdown.properties` to associate the plugin name with our plugin class

carterkozak (approved):
Fantastic, thanks @ferozco!

PR #16, opened 2019-11-25
CreateMetricsManifestTask avoids logging noisy stack traces

carterkozak:
Previously distribution projects which did not produce metrics
resulted in exceptions being harmelssly logged due to attempting
to open files that did not exist.
Now we check for existance first.


ferozco:
👍

PR #17, opened 2019-11-26
Metric manifest contains source of all schemas 

ferozco:
## Before this PR
The metric manifest was a flat list of metric schema that did not provide any way to identify the source of metrics

## After this PR
==COMMIT_MSG==
Generated Metric manifest is now a map of GAVs to lists of metric schema
==COMMIT_MSG==


carterkozak:
Should we put maven coordinates, and potentially vcs URI in the existing options map? Either way I think we should clear the options before we publish data since the options we use in the definition are only for local consumption.

Using the maven coordinate could go sideways if we publish two packages with different classifiers (e.g. tests) where both use metric-schema. I'd like to associate the repo link with our coordinates, I think the repo name will provide a better section header in generated documentation than the group or artifact name. I've seen a few legacy places that duplicate groups across repositories, so it may not be sufficient to group on those. Thoughts?

ferozco:
The metrics manifest is not embedded into published JARs. It is only available locally as an input to the markdown task so i'm not too worried about reverting things down the line if they aren't working out.

I'd like to keep maven coordinates and VCS outside of metric schema for now since that information is already available in the POM of the JAR. Happy to revisit if we decide to start publishing metric-schema as standalone artifacts thoughl

metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
```diff
@@ -25,15 +25,17 @@
...
 /** {@link MarkdownRenderer} consumes consolidated metric schemas from a distribution to produce metrics in markdown. */
 public final class MarkdownRenderer {
 
     /** Returns rendered markdown based on the provided schemas. */
-    public static String render(List<MetricSchema> schemas) {
+    public static String render(Map<String, List<MetricSchema>> schemas) {
         StringBuilder buffer = new StringBuilder();
         buffer.append("# Metrics\n");
-        namespaces(schemas).forEach(namespace -> render(namespace, buffer));
+        // TODO(forozco): use metric-schema provenance in markdown generation
```
carterkozak, line 37: :+1:

carterkozak (approved):
Makes sense, sorry for my confusion :-)

PR #22, opened 2019-12-02
Metric markdown provenance

carterkozak:
Metrics from the local project (as determined by the artifact group)
are listed above those inherited from other projects.

## Before this PR
No way to tell where a metric was defined.

## After this PR
==COMMIT_MSG==
Metric markdown provides provenance. Metrics from the local project (as determined by the artifact group) are listed above those inherited from other projects.
==COMMIT_MSG==

### DIFF
--- a/metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/Section.java
+++ b/metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/Section.java
@@ -18,7 +18,7 @@ package com.palantir.metric.schema.markdown;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.palantir.metric.schema.MetricNamespace;
+import java.util.List;
 import org.immutables.value.Value;
 
 @Value.Immutable
@@ -27,15 +27,15 @@ import org.immutables.value.Value;
         overshadowImplementation = true,
         jdkOnly = true,
         get = {"get*", "is*"})
-@JsonDeserialize(as = ImmutableNamespace.class)
-@JsonSerialize(as = ImmutableNamespace.class)
-interface Namespace {
+@JsonDeserialize(as = ImmutableSection.class)
+@JsonSerialize(as = ImmutableSection.class)
+interface Section {
 
-    String name();
+    String sourceCoordinates();
 
-    MetricNamespace definition();
+    List<Namespace> namespaces();
 
-    class Builder extends ImmutableNamespace.Builder {}
+    class Builder extends ImmutableSection.Builder {}
 
     static Builder builder() {
         return new Builder();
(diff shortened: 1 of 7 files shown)


metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/Section.java
```diff
@@ -0,0 +1,43 @@
...
+package com.palantir.metric.schema.markdown;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(
+        visibility = Value.Style.ImplementationVisibility.PACKAGE,
```
ferozco: default is same as source interface so we can remove this

metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/Section.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.palantir.metric.schema.markdown;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(
```
ferozco, line 25: just in general why do we need this?

metric-schema-markdown/src/test/java/com/palantir/metric/schema/markdown/MarkdownRendererTest.java
```diff
@@ -32,25 +32,39 @@
...
+                                        .type(MetricType.METER)
+                                        .docs(Documentation.of("metric docs"))
+                                        .build())
+                                .build())
                         .build())));
-        assertThat(markdown).isEqualTo("# Metrics\n\n"
-                + "## namespace\n"
+        assertThat(markdown).isEqualTo("# Metrics\n"
+                + "\n"
+                + "## `com.palantir:test:1.0.0`\n"
```
ferozco: I think I prefer the pretty printed name as the title. Could we make the full GAV part of the description to the section

metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/Section.java
carterkozak: Matches standard style we use elsewhere without the additional dependency.

metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/Section.java
carterkozak: :+1:

metric-schema-markdown/src/test/java/com/palantir/metric/schema/markdown/MarkdownRendererTest.java
carterkozak: Yep, that's what the previous commit did, we can revert https://github.com/palantir/metric-schema/pull/22/commits/2fde26244e074006b5d9e3688a577b2210b975be

ferozco (approved):

PR #32, opened 2019-12-05
Exclude version in generated markdown

ferozco:
## After this PR
==COMMIT_MSG==
Exclude version of artifact in generated markdown
==COMMIT_MSG==


metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
```diff
@@ -113,6 +113,12 @@ private static String getName(String coordinate) {
         return results.size() >= 2 ? results.get(1) : coordinate;
     }
 
+    private static String getGroupArtifact(String coordinate) {
+        int lastIndex = coordinate.lastIndexOf(':');
+        int firstIndex = coordinate.indexOf(':');
+        return lastIndex != firstIndex ? coordinate.substring(0, lastIndex): coordinate;
```
carterkozak: Do these ever contain a classifier? e.g. `com.palantir.foo:bar:1.2.3:test`
Should we split on `:`, remove the third group, and join the results

carterkozak (approved):
This will nicely resolve confusion around `$projectVersion` as well, thanks @ferozco!

metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
ferozco: The input is always a GAV so I think its alright

PR #37, opened 2019-12-06
mark all tasks as cacheable

### LINKED ISSUE #30: Gradle tasks should support the build cache
Ideally including remote build caches.

ferozco:
Closes #30 

## After this PR
==COMMIT_MSG==
Mark all tasks as cacheable
==COMMIT_MSG==

## Possible downsides?
If the input/outputs are not correctly the caching may break some people

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CreateMetricsManifestTask.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CreateMetricsManifestTask.java
@@ -43,12 +43,14 @@ import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.provider.Property;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.CacheableTask;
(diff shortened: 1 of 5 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CreateMetricsManifestTask.java
```diff
@@ -43,12 +43,14 @@
...
 import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Classpath;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
 
+@CacheableTask
 public class CreateMetricsManifestTask extends DefaultTask {
```
carterkozak, line 54: Do we need to do anything with @InputFile sensitivity? (I'm pattern matching on the conjure gradle plugin, I don't know much about how these things work!)

carterkozak (approved):

PR #43, opened 2019-12-12
Fix markdown rendering sort order to prioritize local metrics

carterkozak:
Fixes a bug in the markdown renderer which compared artifact
groups to full coordinates.
In the future we shouldn't validate using projects with names
beginning with 'a'. oops!

## After this PR
==COMMIT_MSG==
Fix markdown rendering sort order to prioritize local metrics
==COMMIT_MSG==

## Possible downsides?
none.

### DIFF
--- a/metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
+++ b/metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
@@ -142,23 +142,24 @@ public final class MarkdownRenderer {
      * Comparator which prefers sections from the local projects group. Metrics defined in the local project should be
      * rendered first.
      */
-    private static final class CoordinateComparator implements Comparator<String> {
+    @VisibleForTesting
+    static final class CoordinateComparator implements Comparator<String> {
 
-        private final String localCoordinate;
+        private final String localGroup;
 
         CoordinateComparator(String localCoordinate) {
-            this.localCoordinate = localCoordinate;
+            this.localGroup = getGroup(localCoordinate);
         }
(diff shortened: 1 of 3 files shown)


metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
```diff
@@ -142,23 +142,24 @@ static String getGroup(String coordinate) {
...
         CoordinateComparator(String localCoordinate) {
-            this.localCoordinate = localCoordinate;
+            this.localGroup = getGroup(localCoordinate);
         }
 
         @Override
         public int compare(String first, String second) {
             String firstGroup = getGroup(first);
             String secondGroup = getGroup(second);
-            if (Objects.equals(firstGroup, localCoordinate) || Objects.equals(secondGroup, localCoordinate)) {
```
carterkozak, line 157: :man_facepalming:

metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
```diff
@@ -142,23 +142,24 @@ static String getGroup(String coordinate) {
...
+            if (Objects.equals(firstGroup, localGroup) || Objects.equals(secondGroup, localGroup)) {
                 if (Objects.equals(firstGroup, secondGroup)) {
                     return 0;
```
carterkozak: in this case I think we actually want to fall through... oof.

ferozco (approved):

ferozco:
🚂

PR #50, opened 2019-12-13
Generated utilities use TaggedMetricRegistry.registerWithReplacement

carterkozak:
## Before this PR
This prevents non-obvious resource leaks where attmpts to
re-register gauges are silently ignored while the registry holds
a reference to the original gauge.

## After this PR
==COMMIT_MSG==
Generated utilities use TaggedMetricRegistry.registerWithReplacement

This prevents non-obvious resource leaks where attmpts to
re-register gauges are silently ignored while the registry holds
a reference to the original gauge.

The `registerWithReplacement` api was added in Tritium `0.16.1`, a dependency
upgrade may be required for generated code to compile.
==COMMIT_MSG==

## Possible downsides?
The gradle plugin adds a dependency on the current version of Tritium specified by the project, and doesn't force upgrades. Updated generators may produce code that doesn't compile until Tritium is updated.


iamdanfox:
To address that possible downside - couldn't this gradle plugin just add a dependency on tritium-registry 0.16.1 and then rely on gradle's regular constraint logic to dial it up as appropriate?

carterkozak:
That's a good idea. I wasn't sure how that would interact with GCV but I'll give it a shot!

carterkozak:
This feels a lot cleaner, thanks for the suggestion @iamdanfox!

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -39,6 +39,9 @@
     public static final String GENERATE_METRICS_MARKDOWN = "generateMetricsMarkdown";
     public static final String CREATE_METRICS_MANIFEST = "createMetricsManifest";
 
+    private static final String SAFE_LOGGING_VERSION = "1.12.2";
+    private static final String TRITIUM_VERSION = "0.16.2";
```
schlosna: these end up being the minimum compatible versions, correct? can we bump tritium to `0.16.3` to pick up the error-prone annotations downgrade? the error-prone annotations 2.3.4 is causing some downstream issues I'd prefer to head off.

```suggestion
    private static final String TRITIUM_VERSION = "0.16.3";
```

carterkozak: Good call, updated.

ferozco:
I'd prefer to avoid having the gradle plugin automatically provide versions of dependencies since it makes things more explicit and less magical. Excavator should roll this out to everyone with the release notes so if users encounter an issue they'll having all the information to solve it right in front of them.

carterkozak:
Alright, I've reset to the original implementation and restored the release note comment. I don't have a strong opinion on this, but I am generally in favor of less magic (though also in favor of not having to dig into release notes to figure out why projects don't compile, but if our robots work as expected that won't happen).

ferozco:
👍

PR #53, opened 2019-12-17
Upgrade to tritium 0.16.3

schlosna:
## Before this PR
Using tritium 0.16.2 which may impose unnecessary error-prone annotations 2.3.4 transitive dependency (see https://github.com/palantir/tritium/pull/553 which downgrades that minimum version to 2.3.3)

## After this PR
==COMMIT_MSG==
Upgrade to tritium 0.16.3 
==COMMIT_MSG==

## Possible downsides?
version/release churn


carterkozak:
:+1:

PR #54, opened 2019-12-17
Use TaggedMetricRegistry.registerWithReplacement for all gauges

carterkozak:
## Before this PR
Previously on the simple codepath used this method, I missed the
staged builder path.

## After this PR
==COMMIT_MSG==
Use TaggedMetricRegistry.registerWithReplacement for all gauges
==COMMIT_MSG==


schlosna (approved):

ferozco:
👍

PR #93, opened 2020-01-24
correctly set project group and name as inputs to generateMetricMarkdown

ferozco:
## Before this PR
Ran into some issues when rename a project

## After this PR
==COMMIT_MSG==
correctly set project group and name as inputs to `generateMetricMarkdown`
==COMMIT_MSG==

## Possible downsides?
N/A


carterkozak (approved):
thanks!

PR #275, opened 2020-06-08
fix #274: Relax metric name validation

### LINKED ISSUE #274: Relax metric name validation
The metric name validation added in https://github.com/palantir/metric-schema/pull/271 prevents some previously permitted metric names such as `load.1` (system load over previous one minute) produced by tritium, causing [metric-schema upgrade bump to fail](https://github.com/palantir/tritium/pull/741).

We should consider relaxing the naming constraint to not require an alpha character before numeric after a period, e.g. using regex `[a-z][a-zA-Z0-9\-]+\.)*[a-zA-Z0-9\-]+`

```
* What went wrong:
Execution failed for task ':tritium-metrics-jvm:generateMetrics'.
> MetricDefinition names must match pattern: {pattern=([a-z][a-zA-Z0-9\-]+\.)*[a-z][a-zA-Z0-9\-]+, invalidTagName=load.1}
```

carterkozak:
==COMMIT_MSG==
Relax metric name validation
==COMMIT_MSG==

See https://github.com/palantir/tritium/pull/741#issuecomment-639921046

### DIFF
--- a/metric-schema-java/src/main/java/com/palantir/metric/schema/Validator.java
+++ b/metric-schema-java/src/main/java/com/palantir/metric/schema/Validator.java
@@ -24,7 +24,7 @@ import org.apache.commons.lang3.StringUtils;
 final class Validator {
 
     private static final String SHORT_NAME_PATTERN = "[A-Z][a-zA-Z0-9]+";
-    private static final String NAME_SEGMENT_PATTERN = "[a-z][a-zA-Z0-9\\-]+";
+    private static final String NAME_SEGMENT_PATTERN = "[a-z0-9][a-zA-Z0-9\\-]*";
     private static final String NAME_PATTERN = "(" + NAME_SEGMENT_PATTERN + "\\.)*" + NAME_SEGMENT_PATTERN;
     private static final Pattern NAME_PREDICATE = Pattern.compile(NAME_PATTERN);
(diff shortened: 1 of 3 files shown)


iamdanfox (approved):

metric-schema-java/src/main/java/com/palantir/metric/schema/Validator.java
```diff
@@ -24,7 +24,7 @@
 final class Validator {
 
     private static final String SHORT_NAME_PATTERN = "[A-Z][a-zA-Z0-9]+";
-    private static final String NAME_SEGMENT_PATTERN = "[a-z][a-zA-Z0-9\\-]+";
+    private static final String NAME_SEGMENT_PATTERN = "[a-z0-9][a-zA-Z0-9\\-]*";
     private static final String NAME_PATTERN = "(" + NAME_SEGMENT_PATTERN + "\\.)*" + NAME_SEGMENT_PATTERN;
```
schlosna, line 28: curious, do we want to only allow single digit or single character segment only for the last segment?

carterkozak: I think we were overzealous with the initial validation, WC has a few metrics based on status code, I don't think we need to be too prescriptive.

carterkozak:
🤖

PR #282, opened 2020-06-14
Allow last segment of a namespace to start with an uppercase

Alexis-D:
## Before this PR

Namespaces of the form `com.company.ClassName` are disallowed starting in 0.5.4. We recently migrated to metric-schema and metrics of that form prior to the migration. It would be nice to allow such metric names to avoid breaks.

## After this PR
==COMMIT_MSG==
Allow last segment of a namespace to start with an uppercase
==COMMIT_MSG==

cc @carterkozak


ferozco:
👍 Thanks @Alexis-D.

PR #288, opened 2020-06-17
metric-schema embeds libraryName & libraryVersion tags into all metrics

iamdanfox:
## Before this PR

This has the same motivation as https://github.com/palantir/tritium/pull/754, just an alternative implementation approach which is probably easier to roll out.

> In Dialogue, I've found it incredibly beneficial to have the dialogueVersion tag on all my metrics, because when looking at a graph I can tell whether they're on latest or an older version. I've been using this VersionedTaggedMetricRegistry class to achieve this: https://github.com/palantir/dialogue/blob/develop/dialogue-core/src/main/java/com/palantir/dialogue/core/VersionedTaggedMetricRegistry.java. I had intended to generalize this and make the pattern available for other libraries e.g. atlasdb (cc @jkong) and witchcraft.
>
> A more urgent reason to pursue this came up recently with regard to our DD $ spend, as our spend reports are currently broken down by product tag, which makes it hard to see what proportion of the $ spend our various libraries are responsible for. Apparently we can get cost breakdowns made for any given tag, there just isn't a consistent one across libraries yet. (cc @samrogerson)

## After this PR
==COMMIT_MSG==
Generated code now includes a libraryName and libraryVersion tag in all metrics, to help attribute fleet wide cost to particular libraries. This uses a libraryName derived from the gradle `rootProject.name` by default, although this can be overridden on the `generateMetricSchema` task if desired. (Setting it to `null` turns off the tag entirely).
==COMMIT_MSG==

cc @tpetracca who slacked me about this

## Possible downsides?

note - this PR doesn't currently include these `libraryName` or `libraryVersion` tags in the json / markdown.... is this important??

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
@@ -20,8 +20,11 @@ import com.google.common.collect.ImmutableSet;
 import com.palantir.metric.schema.JavaGenerator;
 import com.palantir.metric.schema.JavaGeneratorArgs;
 import java.io.File;
+import java.util.Optional;
 import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.OutputDirectory;
 import org.gradle.api.tasks.SourceTask;
 import org.gradle.api.tasks.TaskAction;
@@ -30,6 +33,16 @@ import org.gradle.util.GFileUtils;
 @CacheableTask
 public class GenerateMetricSchemaTask extends SourceTask {
     private final DirectoryProperty outputDirectory = getProject().getObjects().directoryProperty();
+    private final Property<String> libraryName = getProject().getObjects().property(String.class);
+
+    public GenerateMetricSchemaTask() {
+        libraryName.set(defaultLibraryName());
+    }
+
+    @Input
+    public final Property<String> getLibraryName() {
+        return libraryName;
+    }
 
     @OutputDirectory
     public final DirectoryProperty getOutputDir() {
@@ -45,8 +58,14 @@ public class GenerateMetricSchemaTask extends SourceTask {
         JavaGenerator.generate(JavaGeneratorArgs.builder()
                 .inputs(getSource().getFiles().stream().map(File::toPath).collect(ImmutableSet.toImmutableSet()))
                 .output(output.toPath())
+                .libraryName(Optional.ofNullable(libraryName.getOrNull()))
                 // TODO(forozco): probably want something better
                 .defaultPackageName(getProject().getGroup().toString())
                 .build());
+    }
+
+    private String defaultLibraryName() {
+        String rootProjectName = getProject().getRootProject().getName();
+        return rootProjectName.replaceAll("-root$", "");
     }
 }
--- a/metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGeneratorArgs.java
+++ b/metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGeneratorArgs.java
@@ -16,8 +16,13 @@
 
 package com.palantir.metric.schema;
 
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import java.nio.file.Path;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import org.immutables.value.Value;
 
 @Value.Immutable
@@ -27,6 +32,9 @@ import org.immutables.value.Value;
         jdkOnly = true,
         get = {"get*", "is*"})
 public abstract class JavaGeneratorArgs {
+
+    private static final Predicate<String> LIBRARY_NAME =
+            Pattern.compile("[a-z0-9]+(-[a-z0-9]+)*").asPredicate();
 
     /**
      * Input directory for metric schema files with the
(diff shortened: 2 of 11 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -45,8 +58,14 @@ public final void generate() {
...
                 .output(output.toPath())
+                .libraryName(Optional.ofNullable(libraryName.getOrNull()))
                 // TODO(forozco): probably want something better
                 .defaultPackageName(getProject().getGroup().toString())
                 .build());
     }
+
+    private String defaultLibraryName() {
+        String rootProjectName = getProject().getRootProject().getName();
+        return rootProjectName.replaceAll("-root$", "");
```
ferozco, line 69: I really think its lame to rely on a convention on a seemingly unrelated project to determine the name. Could we require users to specify the field if they wanted to opt in?

metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGeneratorArgs.java
```diff
@@ -40,9 +48,24 @@
...
+    /** Used to add the libraryName tag to all metrics. */
+    abstract Optional<String> libraryName();
+
     /** The default Java package name for generated classes. */
     abstract String defaultPackageName();
 
+    @Value.Check
+    final void check() {
+        libraryName().ifPresent(value -> {
+            Preconditions.checkArgument(
```
ferozco, line 60: I think these restrictions are another example of why the libraryName should be separate from the root project name

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
carterkozak: On the other hand it would be odd if two components from the same repository had different library names, right?

iamdanfox: When we discussed in the standup, it sounded like we wanted to design this so that library authors who are _not_ us would be likely to get it right. I actually want to keep this functionality as an 'on by default' thing otherwise I don't think any metric-schema users apart from us will flip this switch.

Interestingly the main libraries we care about don't actually have the `-root` suffix. I could just delete it and say if you don't like the default you're free to override?
- c-j-r https://github.com/palantir/conjure-java-runtime/blob/develop/settings.gradle#L1
- dialogue https://github.com/palantir/dialogue/blob/develop/settings.gradle#L1
- atlasdb https://github.com/palantir/atlasdb/blob/develop/settings.gradle#L1
- tritium https://github.com/palantir/tritium/blob/develop/settings.gradle#L1

metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGeneratorArgs.java
iamdanfox: I don't quite understand - can you flesh our your concern more?

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
ferozco: > On the other hand it would be odd if two components from the same repository had different library names, right?

It might be odd, but it could also make sense in especially large repos that produce multiple libraries/distributions.

> I actually want to keep this functionality as an 'on by default' thing otherwise I don't think any metric-schema users apart from us will flip this switch.

Agreed that it makes sense for libraries to have this on by default. I disagree that it makes sense for _all_ users of metric-schema to produce a `libraryName`, `libraryVersion` tag. For services, its at best a bit confusing and at worst we're adding unnecessary strain on the logging infrastructure by producing duplicate information.

metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGeneratorArgs.java
ferozco: Sorry I should have said more words. Its a bit surprising that by applying a plugin on some random sub project I am effectively putting constraints on  the root project. Usually plugin application only impacts the current project and its sub projects

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
ferozco: I just prefer these things to be a bit more explicit and I'm willing trade off the cost/risk of asking libraries to opt in vs. producing unnecessary information for services since libraries are quite uncommon and in general we don't encourage/allow just anyone to write them

tpetracca:
and in case anyone was concerned, adding these should not change unique metric cardinality at all, as for a given a service-node this metric cardinality is strictly less than `product`/`productVersion`.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
iamdanfox: @ferozco I'm actually not convinced by those reasons. I hold the on-by-default thing as _really_ valuable because it ensures that libraries written by non-infra folks will appear in our DataDog reports, and I don't see how this will cause any confusion if the tag gets added a bit superfluously to servers. We've added tags to stuff in the past without telling people, I don't see how this is different. Similarly, I don't think adding a few more bytes to each log line is gonna flatten the signals infra but if it does then I'd be ok with flipping it round.

metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGeneratorArgs.java
iamdanfox: If someone fails this validation maybe they will be a little surprised, but (a) I have yet to come across a repo that would fail this validation and (b) it's a default anyway, so they could choose to override it anyway. 

I'm just trying to automatically do the right thing in the 99% of cases, the 1% of special snowflakes can still override it.

carterkozak (approved):

Issue #242, opened 2020-05-12
Allow removing metrics


a10y:
Adding new metrics is a cinch, but removing metrics requires both

* hard-coding metric name in your code
* Plumbing down MetricRegistry/TaggedMetricRegistry to call its `#remove` method directly

We could potentially generate `remove*` methods for different generated metrics similar to the builders that are currently generated, curious if there are other thoughts here.

carterkozak:
What does it mean to remove a metric? Can you describe the outcome you're looking for?

a10y:
I define removing a metric the way `TaggedMetricRegistry#remove` defines it.

This is most useful for gauges, where I'm tracking a particular flow, and after that flow has concluded I'd like to clean up the metric so it stops reporting the last seen value

carterkozak:
How do you ensure the final value has been recorded prior to removal?

a10y:
In my mind the metrics recording and the metric registry are two separate components. If you have control over the recording process, you could ensure that any values you care about get flushed, and then remove that metric.

In practice, I believe the registry generally gets sampled by a background thread at regular intervals, and so there's no guarantee to the client that any specific measurement was ever recorded anyway.

carterkozak:
This has come up a bit around testing as well, in general we've tried to discourage use of gauges and avoid special casing/complexity around them. How common is this pattern?

a10y:
I can look thru internal codebases and see how prevalent it is, given that most products generally shy away from ephemeral/high-cardinality metrics of any kind my gut says maybe not very. If we think it's an anti-pattern then perhaps it doesn't belong in here. I'll poke around and report back.

carterkozak:
Thanks!

a10y:
So at least in AIP, we use ephemeral metrics very widely,  as measurements attached to specific streams and stream processors. When the stream or the processor is removed, the metric lingers currently. For some of these, we can suppress empty series for dashboards, but counters/gauges will continue to show up permanently.

a10y:
Any thoughts on the right thing to do here? cc @ferozco  as well for input

ferozco:
Totally understand the use case within AIP, but I don't think that its a very common pattern and worth adding support to metric-schema.

Could you continue to use the metric schema generated classes, but wrap the underlying metric registry so that it keeps track of metrics registered to it so you can remove them when the processor closes?

PR #293, opened 2020-06-23
Getters for gauge metric names

iamdanfox:
## Before this PR

Occasionally, people need to interact with a metric name in a way not prescribed by metric-schema. For example, in dialogue we have the `DialogueInternalWeakReducingGauge` thing https://github.com/palantir/dialogue/blob/develop/dialogue-core/src/main/java/com/palantir/dialogue/core/DialogueInternalWeakReducingGauge.java#L64 and in WC.

## After this PR
==COMMIT_MSG==
For any gauges defined in metric-schema, there are now methods that allow getting the raw `MetricName` (either via the staged builder or directly)
==COMMIT_MSG==

cc @carterkozak 

## Possible downsides?

- this could open the door for people to start registering metrics in funky ways. My hope is that by keeping this restricted to gauges people won't get _too_ confused.

### DIFF
--- a/changelog/@unreleased/pr-293.v2.yml
+++ b/changelog/@unreleased/pr-293.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: For any gauges defined in metric-schema, there are now methods that
+    allow getting the raw `MetricName` (either via the staged builder or directly)
+  links:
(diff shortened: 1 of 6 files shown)


carterkozak:
This resolves part of #242

carterkozak (approved):
Looks good. Thanks, Dan!

PR #295, opened 2020-06-25
fooMetricName should be static

iamdanfox:
## Before this PR

I added some methods in 0.5.7 so that it was convenient to access MetricNames for gauges, but I've realised that it presents quite a wonky API. For example, I just wanted to get hold of this MetricName, but I've been forced to pass in a made-up tagged metric registry:

```java
    public static final MetricName WORKER_UTILIZATION_MAX_METRIC_NAME =
            UndertowWorkerMetrics.of(new DefaultTaggedMetricRegistry()).utilizationMaxMetricName();
```

## After this PR
==COMMIT_MSG==
The `.fooMetricName()` getter method for gauges with zero tags can now be accessed statically.
==COMMIT_MSG==

## Possible downsides?
- this makes it a little inconsistent with the other factory methods, which are instance methods

### DIFF
--- a/changelog/@unreleased/pr-295.v2.yml
+++ b/changelog/@unreleased/pr-295.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: The `.fooMetricName()` getter method for a gauge `foo` with zero tags
+    can now be accessed statically.
+  links:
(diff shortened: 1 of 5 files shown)


carterkozak (approved):
I was about to make the same change

PR #314, opened 2020-07-23
generateMetricsMarkdown no longer requires `--write-locks`

carterkozak:
The check portion of the task moved to a new `checkMetricsMarkdown`
task.

## Before this PR
Large internal project can update markdown within ten minutes if you get super lucky.

## After this PR
Large internal project can update markdown within a few seconds.

==COMMIT_MSG==
generateMetricsMarkdown no longer requires `--write-locks`
==COMMIT_MSG==

## Possible downsides?
I don't know how to gradle.
I have no idea what I'm doing

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaMarkdownPlugin.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaMarkdownPlugin.java
@@ -34,7 +34,7 @@ public final class MetricSchemaMarkdownPlugin implements Plugin<Project> {
         TaskProvider<CreateMetricsManifestTask> createMetricsManifest =
                 project.getTasks().named(MetricSchemaPlugin.CREATE_METRICS_MANIFEST, CreateMetricsManifestTask.class);
 
-        project.getTasks()
+        TaskProvider<GenerateMetricMarkdownTask> generateMetricsMarkdown = project.getTasks()
                 .register(MetricSchemaPlugin.GENERATE_METRICS_MARKDOWN, GenerateMetricMarkdownTask.class, task -> {
                     task.getManifestFile().set(createMetricsManifest.flatMap(CreateMetricsManifestTask::getOutputFile));
                     task.getOutputFile().set(project.file("metrics.md"));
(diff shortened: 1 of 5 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaMarkdownPlugin.java
```diff
@@ -34,13 +34,21 @@ public void apply(Project project) {
...
-        TaskProvider<GenerateMetricMarkdownTask> generateMetricsMarkdown = project.getTasks()
+        project.getTasks()
                 .register(MetricSchemaPlugin.GENERATE_METRICS_MARKDOWN, GenerateMetricMarkdownTask.class, task -> {
                     task.getManifestFile().set(createMetricsManifest.flatMap(CreateMetricsManifestTask::getOutputFile));
                     task.getOutputFile().set(project.file("metrics.md"));
                     task.dependsOn(createMetricsManifest);
                 });
-        project.getTasks().named("check", check -> check.dependsOn(generateMetricsMarkdown));
+
+        TaskProvider<CheckMetricMarkdownTask> checkMetricsMarkdown = project.getTasks()
```
ferozco: I'm not sure this will work correctly since `checkMetricsMarkdown` must run before `generateMetricsMarkdown` to ensure that metric.md is up to date but if we do so then we would not be able to update the metrics.md. We would also loose the neat integration with excavator

carterkozak: I'm not sure I follow, both tasks should be independent, but call the same utility to verify the results. `--write-locks` still triggers the generate command, the difference is the generate command no longer requires `--write-locks`.

ferozco: There is currently a race between `checkMetricMarkdown` and `generateMetricMarkdown`. I think if we add `generateMetricMarkdown.mustRunAfter(checkMetricMarkdown)` then we will eliminate the race while allowing you to call `./gradlew generateMetricMarkdown`

carterkozak: I think the race occurs when two tasks are requested in the same run, e.g. `./gradlew checkMetricsMarkdown generateMetricsMarkdown` or `./gradlew check --write-locks`. In that case, wouldn't we want check to run after generating new data? Otherwise it could both fail and produce the correct output.

carterkozak: From chat:

We're worried about accidentally triggering generateMetricsMarkdown from a check task, passing tests, but silently leaving the project in a dirty state. Reversing the mustRunAfter order allows the check to fail.

ferozco:
👍

carterkozak:
:robot:

PR #348, opened 2020-10-13
CreateMetricsManifestTask is not cached

carterkozak:
We've been unable to write a test to reproduce the failure. It's possible there's an up-to-date failure unrelated to the cache.

## Before this PR
Manifests failed to update in some cases, causing unexpected failures on CI

## After this PR
==COMMIT_MSG==
CreateMetricsManifestTask is not cached and no longer causes confusing errors with false cache hits
==COMMIT_MSG==

## Possible downsides?
caching is great when it works as expected. This task is very fast, so it shouldn't be problematic.


ferozco:
👍

PR #373, opened 2020-11-18
fix: correct cache up-to-dateness for project dependencies

ferozco:
## After this PR
==COMMIT_MSG==
correct cache up-to-dateness for project dependencies
==COMMIT_MSG==

cc @carterkozak

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CompileMetricSchemaTask.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CompileMetricSchemaTask.java
@@ -18,7 +18,6 @@ package com.palantir.metric.schema.gradle;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
-import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.conjure.java.serialization.ObjectMappers;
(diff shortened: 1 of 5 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CompileMetricSchemaTask.java
```diff
@@ -27,32 +26,34 @@
...
+import org.gradle.api.tasks.InputFiles;
 import org.gradle.api.tasks.OutputFile;
-import org.gradle.api.tasks.SourceTask;
+import org.gradle.api.tasks.PathSensitive;
+import org.gradle.api.tasks.PathSensitivity;
 import org.gradle.api.tasks.TaskAction;
 
 @CacheableTask
-public class CompileMetricSchemaTask extends SourceTask {
+public abstract class CompileMetricSchemaTask extends DefaultTask {
```
ferozco, line 40: want to avoid using a source task so that we do not skip when source is empty

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CompileMetricSchemaTask.java
```diff
@@ -27,32 +26,34 @@
...
     private static final ObjectReader reader = ObjectMappers.withDefaultModules(new ObjectMapper(new YAMLFactory()))
             .readerFor(MetricSchema.class);
-    private static final ObjectWriter writer =
```
robert3005, line 40: why this?

ferozco: gotta save that object mapper initialization cost

robert3005 (approved):

PR #376, opened 2020-11-25
Fix generateMetrics build cache behaviour

iamdanfox:
## Before this PR

I have been looking at a comparative build scan of a local build vs a CI build, and saw this heartbreaking cache miss:

image

https://gradle.p.b/c/m2khsesqbx2jw/kbk7v6omaei5k/task-inputs?expanded=WyJuejY2Z3RvdTZyNmtjLXNvdXJjZSIsImNwNGxucXJqYXU3d2Etc291cmNlIl0#nz66gtou6r6kc-cacheability

As you can see, we were _so close_ to getting it perfect, but the `/Volumes/git/container-vuln-scanner` path vs `/home/circleci/container-vuln-scanner` thing wrecked us.

_cc @jmcampanini here's another one_

Also jokes that @carterkozak's sixth sense spotted this bug on the original PR: https://github.com/palantir/metric-schema/pull/37#discussion_r354928914

## After this PR
==COMMIT_MSG==
The `generateMetrics` task can now get hits from the gradle build cache irrespective of where the repo is checked out.  Previously you'd only see `FROM-CACHE` hits in the exact checkout path that the original build ran.
==COMMIT_MSG==

I also checked the other tasks annotated with `@CacheableTask`, and none of the others have this problem because they extend DefaultTask so don't have this sneaky implict inherited InputFiles thingy.

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
@@ -22,10 +22,15 @@ import com.palantir.metric.schema.JavaGeneratorArgs;
 import java.io.File;
 import java.util.Optional;
 import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.file.FileTree;
 import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.CacheableTask;
 import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputFiles;
 import org.gradle.api.tasks.OutputDirectory;
+import org.gradle.api.tasks.PathSensitive;
+import org.gradle.api.tasks.PathSensitivity;
+import org.gradle.api.tasks.SkipWhenEmpty;
 import org.gradle.api.tasks.SourceTask;
 import org.gradle.api.tasks.TaskAction;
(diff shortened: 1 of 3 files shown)


robert3005 (approved):

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -39,6 +44,14 @@ public GenerateMetricSchemaTask() {
         libraryName.set(defaultLibraryName());
     }
 
+    @Override
+    @InputFiles
+    @SkipWhenEmpty
+    @PathSensitive(PathSensitivity.RELATIVE)
+    public final FileTree getSource() {
+        return super.getSource();
+    }
```
iamdanfox, line 53: this is what Gradle's SourceTask has baked in:

```java
    @InputFiles
    @SkipWhenEmpty
    @PathSensitive(PathSensitivity.ABSOLUTE)
    public FileTree getSource() {
        return this.sourceFiles.getAsFileTree().matching(this.patternSet);
    }
```

robert3005: yup, absolute is the safe thing to do but it's unnecessary in sane environments

PR #380, opened 2020-12-12
Annotate methods which return values with @CheckReturnValue

mglazer:
This ensures that someone doesn't forget to do:

```
metrics.meter("foo")
```

And assume that works, this ensures that they also use the
return value to call `.mark()`.

## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-380.v2.yml
+++ b/changelog/@unreleased/pr-380.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Metric methods which return a value are now annotated with `@CheckReturnValue`
+    to ensure proper usage.
+  links:
+  - https://github.com/palantir/metric-schema/pull/380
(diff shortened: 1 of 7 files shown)


carterkozak:
Great idea! Thanks, @mglazer :-)

I think we should annotate the builder interface methods as well for the same reason -- in fact the gauge name accessors wouldn't hurt to annotate either.
We may want to update `MetricSchemaPlugin.configureProjectDependencies` to explicitly add an `api` dependency on `com.google.errorprone:error_prone_annotations`. safe-logging preconditions provides it for now, but it makes maintenance easier if we list the dependencies we use explicitly in case of changes.

mglazer:
By builder interface you mean the staged builder? If so fixed.

Also added the project dep, thanks.

carterkozak (approved):

PR #407, opened 2021-03-09
Feature: add support for constants in metric-schema

### LINKED ISSUE #394: Support tags with pre-defined bounded values (i.e. enum tags)
## What happened

I just defined the following metric using metric-schema but there were concerns at CR time that folks would misuse the `reason` tag and start putting unbounded stuff in it.

```yml
 task.noop:
  type: counter
  tags: [reason]
```

A possible solution on my end here is to just define an enum of possible reasons (see below), but I really don't want to maintain my own wrapper around the nice metric-schema generated class. 

```java
public enum TaskNoOpReason {
    FAILURE_TO_ACQUIRE_LOCK,
    NO_LONGER_IN_THE_STORE,
    MISMATCHED_PROCESSING_TIME,
    UNSUPPORTED_CANDIDATE_TYPE
}
```

## Proposal

So the proposal is: what about extending the metric-schema yml format so that I could define the known options upfront somehow, e.g.:

```yml
     task.noop:
        type: counter
        tags: 
        - reason: [failure-to-acquire-lock, no-longer-in-the-store, mismatched-processing-time, unsupported-candidate-type]
```

(I'd still want to support the existing syntax to avoid imposing this on everyone)

ferozco:
## Before this PR
Closes #394.

## After this PR
==COMMIT_MSG==
add support for constants in metric-schema
==COMMIT_MSG==

In the end the flexibility of representing constant tag values as enums won out over the safety of branching staged builders although I do agree it would be unfortunate if folks started to rely on the generated enums to determine business logic.

### DIFF
--- a/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
+++ b/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
@@ -79,6 +79,26 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
         """.stripIndent()
     }
 
+    def 'fails on invalid schema'() {
+        when:
+        file('src/main/metrics/metrics.yml') << """
+        namespaces:
+          server:
+            docs: General web server metrics.
+            metrics:
+              response.size:
+                type: histogram
+                docs: A histogram of the number of bytes written into the response.
+                tags: [service_name]
+                values:
+                  endpoint: [foo]
+        """.stripIndent()
+
+        then:
+        def result = runTasksWithFailure('compileMetricSchema')
+        Throwables.getRootCause(result.getFailure()).getMessage().contains("metric 'response.size' in namespace 'server' has values [endpoint]")
+    }
+
     def 'generates java'() {
         when:
         file('src/main/metrics/metrics.yml') << METRICS
--- a/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
+++ b/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
@@ -90,7 +90,7 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
                 type: histogram
                 docs: A histogram of the number of bytes written into the response.
                 tags:
-                  - tagName: service_name
+                  - tagName: serviceName
                     values: [foo]
         """.stripIndent()
 
@@ -136,7 +136,7 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
 
         then:
         def result = runTasksWithFailure('classes')
-        result.wasExecuted(':generateMetrics')
+        result.wasExecuted(':compileMetricSchema')
         Throwables.getRootCause(result.getFailure()).getMessage().contains("tags must match pattern")
     }
 
@@ -152,8 +152,8 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
 
         then:
         def result = runTasksWithFailure('classes')
-        result.wasExecuted(':generateMetrics')
-        Throwables.getRootCause(result.getFailure()).getMessage().contains("MetricDefinition is required")
+        result.wasExecuted(':compileMetricSchema')
+        Throwables.getRootCause(result.getFailure()).getMessage().contains("null value in entry: my.custom.metric=null")
     }
 
     def 'embeds metrics into jar'() {
--- a/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
+++ b/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
@@ -90,7 +90,7 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
                 type: histogram
                 docs: A histogram of the number of bytes written into the response.
                 tags:
-                  - tagName: serviceName
+                  - name: serviceName
                     values: [foo]
         """.stripIndent()
 
--- a/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
+++ b/gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
@@ -79,7 +79,7 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
         """.stripIndent()
     }
 
-    def 'fails on invalid schema'() {
+    def 'handles shorthand schema'() {
         when:
         file('src/main/metrics/metrics.yml') << """
         namespaces:
@@ -89,14 +89,13 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
               response.size:
                 type: histogram
                 docs: A histogram of the number of bytes written into the response.
-                tags: [service_name]
-                values:
-                  endpoint: [foo]
+                tags:
+                  - tagName: service_name
+                    values: [foo]
         """.stripIndent()
 
         then:
-        def result = runTasksWithFailure('compileMetricSchema')
-        Throwables.getRootCause(result.getFailure()).getMessage().contains("metric 'response.size' in namespace 'server' has values [endpoint]")
+        def result = runTasksSuccessfully('compileMetricSchema')
     }
 
     def 'generates java'() {
--- a/metric-schema-api/src/main/conjure/metric-schema-api.yml
+++ b/metric-schema-api/src/main/conjure/metric-schema-api.yml
@@ -23,7 +23,7 @@ types:
         fields:
           type: MetricType
           tags: set<string>
-          values: map<string, set<string>>
+          values: map<string, set<TagValue>>
           docs: Documentation
       MetricType:
         values:
@@ -35,3 +35,6 @@ types:
       Documentation:
         docs: Documentation describing an associated component. Markdown syntax may be used.
         alias: string
+      TagValue:
+        fields:
+          value: string
--- a/metric-schema-api/src/main/conjure/metric-schema-api.yml
+++ b/metric-schema-api/src/main/conjure/metric-schema-api.yml
@@ -23,7 +23,7 @@ types:
         fields:
           type: MetricType
           tags: set<string>
-          values: map<string, set<TagValue>>
+          tagDefinitions: set<TagDefinition>
           docs: Documentation
       MetricType:
         values:
@@ -35,6 +35,7 @@ types:
       Documentation:
         docs: Documentation describing an associated component. Markdown syntax may be used.
         alias: string
-      TagValue:
+      TagDefinition:
         fields:
-          value: string
+          name: string
+          values: set<string>
--- a/metric-schema-api/src/main/conjure/metric-schema-api.yml
+++ b/metric-schema-api/src/main/conjure/metric-schema-api.yml
@@ -23,7 +23,7 @@ types:
         fields:
           type: MetricType
           tags: set<string>
-          tagDefinitions: set<TagDefinition>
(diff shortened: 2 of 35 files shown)


ferozco:
You can see example adoption internally: https://g.p.b/f/hype/pull/1133

carterkozak (commented):
Enums are reasonable, we may need to be more careful about avoiding collisions because many metrics share common prefixes, e.g. metric `foo` with tag `bar-baz` and metric `foo.bar` with tag `baz`. Perhaps using private static nested classes based on the metric base-name? Alternatively we can check for collisions with some fallback, but that means writing two codepaths.

At this point we should provide a friendly user-facing schema like conjure rather than forcing users to configure multiple fields. We can update the user-facing yaml parser to handle both single string elements (the current released syntax) as well as tag-name to set of allowed values in order to keep things both legible, and retain the user-provided order:

```yaml
namespaces:
  monitors:
    docs: General web server metrics.
    metrics:
      processing:
        docs: Measures number of installations that were processed
        type: meter
        tags:
          - result: [success, failure]
          - type
          - documentedTag:
            values: [a, b]
            docs: This tag has documentation!
```

This allows us to transform our data into a backwards compatible conjure-based IR, deprecating our existing `tags: set<string>`, and adding a new field which takes a union of tag definitions of either name, or name,values tuple. Old consumers will continue to read the fully populated tags field, while newer code can parse the tagDefinitions when they exist. If we allow tags to be defined, we should be able to provide docs for each tag value. Perhaps require documentation?

We should update the markdown generator to take advantage of this data as well!

ferozco:
> Enums are reasonable, we may need to be more careful about avoiding collisions because many metrics share common prefixes, e.g. metric foo with tag bar-baz and metric foo.bar with tag baz. Perhaps using private static nested classes based on the metric base-name? Alternatively we can check for collisions with some fallback, but that means writing two codepaths.

Good catch! Where would you nest the classes? Unfortunately we already branch if a metric has a single tag so we're not guaranteed to have a builder class. I'll improve the enum name generation to avoid collisions.

>At this point we should provide a friendly user-facing schema like conjure rather than forcing users to configure multiple fields. 

@carterkozak I do like the structure that you propose but I'm a bit worried about adding custom parser logic since the current Conjure-defined yml structure is relied upon by the golang implementation and it would make it harder to keep the two in sync. We _could_ go down the route of publishing a CLI for parsing the yaml but I'd like to keep this simple for as long as possible so open to alternatives. Maybe @nmiyake has some thoughts.

carterkozak:
> Where would you nest the classes?

We could generate a bested [MetricName]Tags nested class for each metric which encapsulates enums, otherwise use an underscore in the name. Both do effectively the same thing.

> I do like the structure that you propose but I'm a bit worried about adding custom parser logic since the current Conjure-defined yml structure is relied upon by the golang implementation and it would make it harder to keep the two in sync.

I’d rather make the infra a little more complicated than force every dev using the tool to understand additional complexity. Reimplementing the parser logic in go shouldn’t be too bad, this api is much simpler than conjure.

ferozco:
@carterkozak think this is ready for another look

metric-schema-lang/src/main/java/com/palantir/metric/schema/lang/TagDefinition.java
```diff
@@ -0,0 +1,39 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Set;
+import org.immutables.value.Value.Immutable;
+
+@Immutable
+public interface TagDefinition {
+    String tagName();
+
+    Set<String> values();
```
carterkozak: It would be helpful if we had the option to describe each tag, perhaps even each predefined tag value. What do you think?

ferozco: I could see it being a nice follow up :)

carterkozak: Of course, however it does have some bearing on the `metric-schema-api.yml` change to allow us to support more expressive syntax without breaking consumers.

ferozco: Sorry, I may have been overly terse. I believe we can provide that functionality in a forward compat way by doing the same thing that this PR does with tags. I.e Supporting the following:

```
namespaces:
  monitors:
    docs: General web server metrics.
    metrics:
      processing:
        docs: Measures number of installations that were processed
        type: meter
        tags:
          - result: [success, failure]
          - type
          - documentedTag:
              values:
                - value: a
                  docs: Some docs about tag
                - b
              docs: This tag has documentation!
```

carterkozak: Right, the `-lang` component should be straightforward, but we need to ensure the IR has a way to represent new data without deprecating a tags field and adding another each time we want to add a docs field.

ferozco: okiedokes, I misunderstood. Should be fixed up

metric-schema-api/src/main/conjure/metric-schema-api.yml
```diff
@@ -23,6 +23,7 @@ types:
         fields:
           type: MetricType
           tags: set<string>
+          values: map<string, set<TagValue>>
```
carterkozak: Should this be a list/set e.g. `tagDefinitions: set<TagDefinition>` in order to support additional data on the tag itself (tag level documentation and future types of constraints, for example)?

That would allow us to deprecate and eventually remove the existing `tags` field. In the meantime we can continue to add all tag names to that set to avoid breaking compatibility.

gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
```diff
@@ -79,6 +79,25 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
...
+        file('src/main/metrics/metrics.yml') << """
+        namespaces:
+          server:
+            docs: General web server metrics.
+            metrics:
+              response.size:
+                type: histogram
+                docs: A histogram of the number of bytes written into the response.
+                tags:
+                  - tagName: serviceName
```
carterkozak: nit: Thoughts on `name` since `tag` context is provided by the parent set?

ferozco: done

metric-schema-api/src/main/conjure/metric-schema-api.yml
ferozco: I've made the change as you've requested. I made it so that the Java generator ignores the old field (`tags`) since it should only ever be invoked with a  version of the compiler that populated `tagDefinitions`. The markdown generator supports both fields since it can be operating on old schemas that have not adopted the new structure

metric-schema-api/src/main/conjure/metric-schema-api.yml
```diff
@@ -35,6 +35,7 @@ types:
       Documentation:
         docs: Documentation describing an associated component. Markdown syntax may be used.
         alias: string
-      TagValue:
+      TagDefinition:
         fields:
-          value: string
+          name: string
+          values: set<string>
```
carterkozak: The `TagValue` type grated us some additional flexibility. Any reason not to include both, using `set<TagValue>` here?

metric-schema-api/src/main/conjure/metric-schema-api.yml
carterkozak: :+1:

metric-schema-java/src/integrationInput/java/com/palantir/test/MonitorsMetrics.java
```diff
@@ -0,0 +1,112 @@
...
+
+        @Override
+        public ProcessingBuilder result(Processing_Result result) {
+            Preconditions.checkState(
+                    this.result == null,
+                    "TagDefinition{name: result, values: [TagValue{value: success}, TagValue{value: failure}]} is already set");
+            this.result =
+                    Preconditions.checkNotNull(
+                            result,
+                            "TagDefinition{name: result, values: [TagValue{value: success}, TagValue{value: failure}]} is required");
```
carterkozak: I suspect we want `"result is already set"` and `"result is required"`

ferozco: done

metric-schema-java/src/integrationInput/java/com/palantir/test/VisibilityMetrics.java
```diff
@@ -39,7 +39,7 @@ Counter test() {
 
     /** Tagged gauge metric. */
     @CheckReturnValue
-    ComplexBuilderFooStage complex() {
+    ComplexBuilderBarStage complex() {
```
carterkozak: I don't think we should reorder builder stages, this would require a great deal of manual effort for large internal projects that heavily utilize metric-schema.

Originally I think we agreed to preserve tag definition order from the input. I suspect this is the result of a non-linked set.

metric-schema-lang/src/main/java/com/palantir/metric/schema/lang/Validator.java
```diff
@@ -35,7 +42,7 @@ static void validate(MetricSchema schema) {
         schema.getNamespaces().forEach(Validator::validateNamespace);
     }
 
-    private static void validateNamespace(String namespace, MetricNamespace namespaceValue) {
+    private static void validateNamespace(String namespace, com.palantir.metric.schema.MetricNamespace namespaceValue) {
```
carterkozak: I'm surprised that the `UnnecessarilyQualified` errorprone check isn't complaining about `com.palantir.metric.schema.MetricNamespace` when we have an import on line 22.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -208,22 +250,27 @@ private static void generateMetricFactoryBuilder(
                 .addModifiers(visibility.apply())
                 .addMethods(abstractBuildMethods)
                 .build());
-        ImmutableList<String> tagList = ImmutableList.copyOf(definition.getTags());
+        ImmutableList<TagDefinition> tagList = ImmutableList.copyOf(definition.getTagDefinitions());
```
ferozco: @carterkozak it seems like we had previously relied on the implicit order of the set when generating the builder stages. I'll make it alphabetical based off of the tag name, but this may some impact consumers

metric-schema-lang/src/main/java/com/palantir/metric/schema/lang/Validator.java
ferozco: odd, it might be because there is another class with the same name in the same package. Fixed it anyways

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
carterkozak: We discussed tag order when we initially implemented metric-schema :-)
Unlike conjure, we want to reflect the input because it may be meaningful, and changes only impact code in the consumer project. Stable order doesn't buy us anything and costs legibility.

Take our metrics with `service` and `endpoint` tags for example, you'd expect to start with `service` and build specificity toward the `endpoint` tag. Similarly, if you have a series of metrics with both `service` and `endpoint` tags, you may have additional tags on some metrics, e.g. `framework` or `channel-name`. imo it's jarring if these break a common pattern.

metric-schema-lang/src/main/java/com/palantir/metric/schema/lang/Validator.java
carterkozak: It may be cleaner to work with the code if we prefix or suffix the lang classes in some way e.g. `LangMetricNamespace`.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
ferozco: Ok, I've updated the api and made it explicit that we will maintain the order of tags.

carterkozak (approved):
lgtm

PR #433, opened 2021-04-01
Fix publish

CRogers:
## Before this PR
We publish invalid poms for the conjure-json.

## After this PR
==COMMIT_MSG==
Fixed publishing.
==COMMIT_MSG==


carterkozak (approved):
tyvm

PR #443, opened 2021-04-12
Improvement: Inline single value tags

ferozco:
## Before this PR
Although a tag was declared to have a single value, users were still forced to provide the value when constructing the metric object

## After this PR
==COMMIT_MSG==
Inline single value tags
==COMMIT_MSG==


carterkozak (approved):
thanks!

PR #464, opened 2021-05-28
Ignore local metrics in discovered metrics

pkoenig10:
We have a project with the following setup:
```gradle
// foo-lib/build.gradle
dependencies {
    implementation project(':foo-server')
}

// foo-server/build.gradle
dependencies {
    runtimeOnly project(':foo-lib')
}
```

`foo-server:createMetricsManifest` fails with `Multiple entries with same key` because the `foo-server` metrics are both local and discovered.

Ideally we would pull out the `foo-server` interfaces needed by `foo-lib` to a separate library, but this is a non-trivial undertaking. It seems reasonable for metric-schema to properly support these types of projects.


carterkozak (approved):
I'm surprised our large internal friends haven't hit this given how heavily they rely on the circular runtime dep pattern. I think they use nearly-codeless distribution modules which don't directly declare metrics themselves.

Change looks good to me, projects would ideally be structured better, but metric-schema shouldn't be an enforcement mechanism to prevent it.

pkoenig10:
> I think they use nearly-codeless distribution modules which don't directly declare metrics themselves.

Yeah this is what we currently do, but I'm trying to do some clean up and would like to combine our server and distribution modules.

PR #470, opened 2021-07-13
Require jdk11 and format java code with Goethe

carterkozak:
## Before this PR
duplication

## After this PR
==COMMIT_MSG==
Require jdk11 and format java code with Goethe
==COMMIT_MSG==

## Possible downsides?
jdk11


metric-schema-java/src/main/java/com/palantir/metric/schema/JavaGenerator.java
```diff
@@ -18,16 +18,21 @@
...
-    /** Specifies under which package Java classes should be generated. */
+    /**
+     * Specifies under which package Java classes should be generated.
+     */
     private static final String JAVA_PACKAGE = "javaPackage";
 
-    /** Specifies visibility of generated Utility class. Defaults to public */
+    /**
+     * Specifies visibility of generated Utility class. Defaults to public
+     */
```
carterkozak, line 35: I wish we could retain the nice one-line javadoc with pjf, but that's a nit for another day :-)

ferozco (approved):

PR #480, opened 2021-07-22
Feature: Support nested directories in metric schema source set

ferozco:
## After this PR
==COMMIT_MSG==
Support nested directories in metric schema source set
==COMMIT_MSG==


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -113,7 +113,7 @@ private static SourceDirectorySet createSourceSet(Project project) {
         SourceDirectorySet sourceSet =
                 project.getObjects().sourceDirectorySet("metricSchema", "Metric Schema source set");
         sourceSet.srcDir("src/main/metrics");
-        sourceSet.include("*.yml");
+        sourceSet.include("**/*.yml");
```
carterkozak, line 116: Suppose it works for conjure, might as well support it here!

carterkozak (approved):

PR #517, opened 2021-09-10
Maintain sourceDirs and generatedSourceDirs mutability

TheKeveloper:
## Before this PR
The current implementation of the MetricSchemaPlugin sets sourceDirs and generatedSourceDirs to be immutable sets. However, other gradle plugins do not expect these sets to be immutable. Making them immutable can break other plugins that add to sourceDirs or generatedSourceDirs in a mutable fashion.

## After this PR
==COMMIT_MSG==
Instead of setting sourceDirs and generatedSourceDirs to be an ImmutableSet, retrieve the existing respective sets and add to them. This is the commonly accepted method of appending sourceDirs and generatedSourceDirs, and keeps each set mutable to avoid breaking other plugins. 
==COMMIT_MSG==

## Possible downsides?
The MetricsSchemaPlugin now depends on other plugins maintaining the mutability of sourceDirs and generatedSourceDirs.

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
@@ -16,8 +16,6 @@
 
 package com.palantir.metric.schema.gradle;
 
-import com.google.common.collect.ImmutableSet;
-import java.io.File;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/metric-schema](https://github.com/palantir/metric-schema), @TheKeveloper! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

TheKeveloper:
Closing this since it was moved.

PR #625, opened 2022-01-18
Keep sourceDirs and generatedSourceDirs mutable by adding to existing…

schlosna:
Pulling over from https://github.com/palantir/metric-schema/pull/517 onto in-repo branch

## Before this PR
The current implementation of the MetricSchemaPlugin sets sourceDirs and generatedSourceDirs to be immutable sets. However, other gradle plugins do not expect these sets to be immutable. Making them immutable can break other plugins that add to sourceDirs or generatedSourceDirs in a mutable fashion.

## After this PR
==COMMIT_MSG==
Keep sourceDirs and generatedSourceDirs mutable by adding to existing set.

Instead of setting sourceDirs and generatedSourceDirs to be an ImmutableSet, retrieve the existing respective sets and add to them. This is the commonly accepted method of appending sourceDirs and generatedSourceDirs, and keeps each set mutable to avoid breaking other plugins. 
==COMMIT_MSG==

## Possible downsides?
The MetricsSchemaPlugin now depends on other plugins maintaining the mutability of sourceDirs and generatedSourceDirs.

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
@@ -16,8 +16,6 @@
 
 package com.palantir.metric.schema.gradle;
 
-import com.google.common.collect.ImmutableSet;
-import java.io.File;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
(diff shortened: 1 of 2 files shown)


carterkozak (approved):
Not sure how I missed the other PR -- thanks for bringing this up!

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -137,14 +135,8 @@ private static void configureIdea(
...
                 IdeaModule module = idea.getModule();
-                module.setSourceDirs(ImmutableSet.<File>builder()
-                        .addAll(module.getSourceDirs())
-                        .add(outputDir.get().getAsFile())
-                        .build());
-                module.setGeneratedSourceDirs(ImmutableSet.<File>builder()
-                        .addAll(module.getGeneratedSourceDirs())
-                        .add(outputDir.get().getAsFile())
-                        .build());
+                module.getSourceDirs().add(outputDir.get().getAsFile());
```
ferozco, line 138: should we always be defensive and construct a new mutable set with all of the entities?

schlosna: seems reasonable, though gradle assumption seems to be that these sets would be mutable. I ran into this when adding metric-schema to a project with another gradle plugin that appends to `sourceDirs`

```
Caused by: java.lang.UnsupportedOperationException
        at com.google.common.collect.ImmutableCollection.add(ImmutableCollection.java:264)
        at com.palantir.foo.gradle.FooGroovyGradleUtils$_addGeneratedSourceDir_closure2$_closure9.doCall(FooGroovyGradleUtils.groovy:56)
```

Conjure plugin does take the defensive approach via `module.setSourceDirs(mutableSetWithExtraEntry(module.getSourceDirs(), ...`

https://github.com/palantir/gradle-conjure/blob/4ba6e7ba9f72c260a8cfd13dd7a8d9f1e9d8e742/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjurePlugin.java#L589-L607

PR #731, opened 2022-04-05
Generated metric utilities require safe tag values

carterkozak:
==COMMIT_MSG==
Generated metric utilities require safe tag values
==COMMIT_MSG==

### DIFF
--- a/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
+++ b/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.errorprone.annotations.CheckReturnValue;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.Safe;
 import com.palantir.tritium.metrics.registry.MetricName;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import com.squareup.javapoet.ClassName;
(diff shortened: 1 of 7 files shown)


metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -71,8 +73,8 @@ static JavaFile generateUtilityClass(
                             Modifier.STATIC,
                             Modifier.FINAL)
                     .initializer(
-                            "$T.ofNullable($T.class.getPackage().getImplementationVersion()).orElse(\"unknown\")",
-                            Optional.class,
+                            "$T.requireNonNullElse($T.class.getPackage().getImplementationVersion(), \"unknown\")",
+                            Objects.class,
```
carterkozak, line 77: While I was here: avoid unnecessary Optional allocation

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -50,7 +52,7 @@ static JavaFile generateUtilityClass(
                 .addModifiers(visibility.apply(Modifier.FINAL))
                 .addJavadoc(Javadoc.render(metrics.getDocs()))
                 .addMethod(MethodSpec.methodBuilder(ReservedNames.FACTORY_METHOD)
-                        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                        .addModifiers(visibility.apply(Modifier.STATIC))
```
carterkozak, line 55: replaced a couple places which specified the public modifier despite being unnecessary based on the encapsulating classes visibility (no consumer impact, just cleaner generated code using a utility we already have)

jkozlowski (approved):

PR #778, opened 2022-05-10
Add JSR330's Inject annotation when available

westinrm:
## Before this PR
Consumers of metric-schema that use dagger end up manually creating modules that @Provide every generated metric schema type.

## After this PR
==COMMIT_MSG==
Add JSR330's Inject annotation when available
==COMMIT_MSG==

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
@@ -37,6 +37,8 @@ import org.gradle.util.GFileUtils;
 public abstract class GenerateMetricSchemaTask extends DefaultTask {
     private final Property<String> libraryName =
             getProject().getObjects().property(String.class).value(defaultLibraryName());
+    private final Property<Boolean> generateDaggerAnnotations =
+            getProject().getObjects().property(Boolean.class).value(false);
 
     @InputFile
     @PathSensitive(PathSensitivity.RELATIVE)
@@ -45,6 +47,11 @@ public abstract class GenerateMetricSchemaTask extends DefaultTask {
     @Input
     public final Property<String> getLibraryName() {
         return libraryName;
+    }
+
+    @Input
+    public final Property<Boolean> getGenerateDaggerAnnotations() {
+        return generateDaggerAnnotations;
     }
 
     @OutputDirectory
@@ -60,6 +67,7 @@ public abstract class GenerateMetricSchemaTask extends DefaultTask {
                 .input(getInputFile().getAsFile().get().toPath())
                 .output(output.toPath())
                 .libraryName(Optional.ofNullable(libraryName.getOrNull()))
+                .generateDaggerAnnotations(getGenerateDaggerAnnotations().get())
                 // TODO(forozco): probably want something better
                 .defaultPackageName(getProject().getGroup().toString())
                 .build());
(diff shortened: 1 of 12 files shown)


carterkozak (changes requested):
I don't think this library should be aware of dagger, there may be other ways to achieve what you're looking for

pkoenig10:
I don't think this library needs to be aware of Dagger specifically. It just needs to have accessible constructors annotated with the JSR-330 `@Inject` annotation. All popular dependency injection frameworks use JSR-330 annotations.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -37,6 +37,8 @@
 public abstract class GenerateMetricSchemaTask extends DefaultTask {
     private final Property<String> libraryName =
             getProject().getObjects().property(String.class).value(defaultLibraryName());
+    private final Property<Boolean> generateDaggerAnnotations =
```
pkoenig10, line 40:
```suggestion
    private final Property<Boolean> generateInjectAnnotations =
```

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -70,6 +74,17 @@ public void apply(Project project) {
                         .apply(MetricSchemaMarkdownPlugin.class));
     }
 
+    private static boolean hasInjectionFramework(Project project) {
+        Configuration compileClasspath = project.getConfigurations().getByName("compileClasspath");
```
pkoenig10, line 78:
```suggestion
        SourceSet main = project.getExtensions()
                .getByType(JavaPluginExtension.class)
                .getSourceSets()
                .getByName(SourceSet.MAIN_SOURCE_SET_NAME)
        Configuration compileClasspath = project.getConfigurations()
                .getByName(main.getCompileClasspathConfigurationName());
```

metric-schema-java/src/integrationInputWithInject/java/com/palantir/test/ServerMetrics.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String LIBRARY_NAME = "witchcraft";
+
+    private static final String LIBRARY_VERSION =
+            Objects.requireNonNullElse(ServerMetrics.class.getPackage().getImplementationVersion(), "unknown");
+
+    private final TaggedMetricRegistry registry;
+
+    @Inject
+    public ServerMetrics(TaggedMetricRegistry registry) {
+        this.registry = registry;
```
carterkozak, line 26: There are a few odd things going on here:
- new users will be fine, however future removal of jakarta-inject deps may result in compilation failures for folks who manually called the public constructor without realizing that's not recommended.
- The public constructor doesn't validate against nulls, resulting in potential NPEs after initialization

We scoped out a potential features that we haven't had the time to implement in which a namespace could declare tags across all nested metrics (much like library name/version, but user-defined). We'd end up with a builder for the metrics class e.g.

```java
ServerMetrics metrics = ServerMetrics.builder()
    .registry(taggedMetricRegistry)
    .serviceName(service)
    .endpointName(endpoint)
    .build();
Histogram responseSize = metrics.responseSize();
```
Such an improvement wouldn't work very well with ctor-based injection

westinrm: namespace scoped metrics would make this pretty painful.  If that's something that you think will get added I'd say just abandon this PR.  It's definitely not critical, just convenient.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
westinrm: If we want to use that notation should I also change https://github.com/palantir/metric-schema/blob/develop/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java#L78 ?

PR #805, opened 2022-06-11
gradle task configuration avoidance

rzpt:
## Before this PR
We are unnecessarily creating and configuring tasks during gradle configuration. See https://docs.gradle.org/current/userguide/task_configuration_avoidance.html for more details.

## After this PR
==COMMIT_MSG==
Avoid creating and configuring gradle tasks
==COMMIT_MSG==

## Possible downsides?
None.

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
@@ -19,6 +19,7 @@ package com.palantir.metric.schema.gradle;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
+import org.gradle.api.UnknownDomainObjectException;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.file.Directory;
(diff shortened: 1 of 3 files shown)


CRogers (approved):

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -122,17 +124,18 @@ private static void configureJavaSource(Project project, Provider<Directory> out
 
     private static void configureEclipse(Project project, TaskProvider<? extends Task> generateMetrics) {
         project.getPluginManager().withPlugin("eclipse", _plugin -> {
-            Task task = project.getTasks().findByName("eclipseClasspath");
-            if (task != null) {
-                task.dependsOn(generateMetrics);
+            try {
+                project.getTasks().named("eclipseClasspath").configure(t -> t.dependsOn(generateMetrics));
+            } catch (UnknownDomainObjectException e) {
```
CRogers, line 129: Might be a bit clearer to use [`UnknownTaskException`](https://docs.gradle.org/current/javadoc/org/gradle/api/UnknownTaskException.html) as that's what's explicitly called out in the javadoc.

PR #840, opened 2022-08-10
Allow to specify namespace tags.

jkozlowski:
## Before this PR

There are usecases where namespaces group metrics which should have the same tags applied to all of them. Right now the solution is to copy-pasta 🍝  the tags into all the metrics. We can probably do better than that.

TODO:

- [x] toString.
- [x] Pass and assign variables in constructor.
- [x] Generate enums.
- [x] Add namespace tags to metric tags.
- [x] Staged builder.
- [x] Validation.
- [x] Markdown generation.
- [x] README

## After this PR
==COMMIT_MSG==
Allow to specify namespace tags.
==COMMIT_MSG==

### DIFF
--- a/metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
+++ b/metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
@@ -6,8 +6,6 @@ import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.Safe;
 import com.palantir.tritium.metrics.registry.MetricName;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Objects;
 
 /**
@@ -23,21 +21,27 @@ public final class NamespaceTagsMetrics {
 
     private final TaggedMetricRegistry registry;
 
-    private final Map<String, String> tags;
+    private final String noValueTag;
 
-    private NamespaceTagsMetrics(TaggedMetricRegistry registry, Map<String, String> tags) {
+    private final NamespaceTags_LocatorWithMultipleValues locatorWithMultipleValues;
+
+    private NamespaceTagsMetrics(
+            TaggedMetricRegistry registry,
+            String noValueTag,
+            NamespaceTags_LocatorWithMultipleValues locatorWithMultipleValues) {
         this.registry = registry;
-        this.tags = tags;
+        this.noValueTag = noValueTag;
+        this.locatorWithMultipleValues = locatorWithMultipleValues;
     }
 
     public static NamespaceTagsMetrics of(
             TaggedMetricRegistry registry,
             String noValueTag,
             NamespaceTags_LocatorWithMultipleValues locatorWithMultipleValues) {
-        Map<String, String> tags = new HashMap<>();
-        tags.put("noValueTag", noValueTag);
-        tags.put("locatorWithMultipleValues", locatorWithMultipleValues.getValue());
-        return new NamespaceTagsMetrics(Preconditions.checkNotNull(registry, "TaggedMetricRegistry"), tags);
+        return new NamespaceTagsMetrics(
+                Preconditions.checkNotNull(registry, "TaggedMetricRegistry"),
+                Preconditions.checkNotNull(noValueTag, "noValueTag"),
+                Preconditions.checkNotNull(locatorWithMultipleValues, "locatorWithMultipleValues"));
     }
 
     /**
@@ -55,8 +59,9 @@ public final class NamespaceTagsMetrics {
     public Meter more() {
         return registry.meter(MetricName.builder()
                 .safeName("namespace-tags.more")
-                .putAllSafeTags(tags)
                 .putSafeTags("locator", "package:identifier")
+                .putSafeTags("noValueTag", noValueTag)
+                .putSafeTags("locatorWithMultipleValues", locatorWithMultipleValues.getValue())
                 .putSafeTags("otherLocator2", "package:identifier")
                 .putSafeTags("libraryName", LIBRARY_NAME)
                 .putSafeTags("libraryVersion", LIBRARY_VERSION)
@@ -155,8 +160,9 @@ public final class NamespaceTagsMetrics {
         public Meter build() {
             return registry.meter(MetricName.builder()
                     .safeName("namespace-tags.processing")
-                    .putAllSafeTags(tags)
                     .putSafeTags("locator", "package:identifier")
+                    .putSafeTags("noValueTag", noValueTag)
+                    .putSafeTags("locatorWithMultipleValues", locatorWithMultipleValues.getValue())
                     .putSafeTags("result", result.getValue())
                     .putSafeTags("type", type)
                     .putSafeTags("otherLocator", otherLocator.getValue())
--- a/metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
+++ b/metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
@@ -35,7 +35,6 @@ public final class NamespaceTagsMetrics {
             String noValueTag,
             NamespaceTags_LocatorWithMultipleValues locatorWithMultipleValues) {
         Map<String, String> tags = new HashMap<>();
-        tags.put("locator", "package:identifier");
         tags.put("noValueTag", noValueTag);
         tags.put("locatorWithMultipleValues", locatorWithMultipleValues.getValue());
         return new NamespaceTagsMetrics(Preconditions.checkNotNull(registry, "TaggedMetricRegistry"), tags);
@@ -57,6 +56,7 @@ public final class NamespaceTagsMetrics {
         return registry.meter(MetricName.builder()
                 .safeName("namespace-tags.more")
                 .putAllSafeTags(tags)
+                .putSafeTags("locator", "package:identifier")
                 .putSafeTags("otherLocator2", "package:identifier")
                 .putSafeTags("libraryName", LIBRARY_NAME)
                 .putSafeTags("libraryVersion", LIBRARY_VERSION)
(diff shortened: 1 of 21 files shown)


metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
```diff
@@ -0,0 +1,189 @@
...
+        this.registry = registry;
+        this.tags = tags;
+    }
+
+    public static NamespaceTagsMetrics of(
+            TaggedMetricRegistry registry,
+            String noValueTag,
+            NamespaceTags_LocatorWithMultipleValues locatorWithMultipleValues) {
+        Map<String, String> tags = new HashMap<>();
+        tags.put("locator", "package:identifier");
```
jkozlowski: I think we probably don't want to put this in the map, but instead manually shove them everytime right? seems like that's how the other code gets generated, for memory usage?

carterkozak: I might go so far as keeping each of the global tags in an instance field instead of constructing a map in the generated code. Runtime complexity is similar (memory access is likely slightly easier for machines to reason about) but also as a human, the source code makes it more obvious what tags are included with a given metric.

jkozlowski: yeaaaa

jkozlowski: Alright, done.

metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
```diff
@@ -0,0 +1,196 @@
...
+        private String type;
+
+        private Processing_OtherLocator otherLocator;
+
+        @Override
+        public Meter build() {
+            return registry.meter(MetricName.builder()
+                    .safeName("namespace-tags.processing")
+                    .putSafeTags("locator", "package:identifier")
+                    .putSafeTags("noValueTag", noValueTag)
```
jkozlowski, line 222: This fallthrough to outer class is very serendipidous, I'm not sure we've made the inner classes non-static on purpose. I think I'd like to ship with this and maybe refactor at some other time; I guess we're trading memory locality vs memory usage?

carterkozak: I think it was intentional as a way to access the `registry`. The namespaced tag access is no different, I like it

jkozlowski: ah, missed that! perfect then

metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
```diff
@@ -55,8 +59,9 @@ public ProcessingBuilderResultStage processing() {
     public Meter more() {
         return registry.meter(MetricName.builder()
                 .safeName("namespace-tags.more")
-                .putAllSafeTags(tags)
                 .putSafeTags("locator", "package:identifier")
+                .putSafeTags("noValueTag", noValueTag)
+                .putSafeTags("locatorWithMultipleValues", locatorWithMultipleValues.getValue())
```
carterkozak: Thoughts on extracting the string-value when we instantiate this utility object rather than each time a metric is accessed? That way all the namespaced tag fields use the `String` type.

metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public static NamespaceTagsMetrics of(
+            TaggedMetricRegistry registry,
+            String noValueTag,
+            NamespaceTags_LocatorWithMultipleValues locatorWithMultipleValues) {
+        return new NamespaceTagsMetrics(
+                Preconditions.checkNotNull(registry, "TaggedMetricRegistry"),
+                Preconditions.checkNotNull(noValueTag, "noValueTag"),
+                Preconditions.checkNotNull(locatorWithMultipleValues, "locatorWithMultipleValues"));
+    }
```
carterkozak: Not sure if you're already tracking, but I think we should make this a staged builder, otherwise tags with the same type will be easy to order incorrectly (same reason we use staged builders for individual metric builders)

jkozlowski: Yes, just wanted to get the simple thing first. I'll get validation and the staged builder done next. Reason I hesitated is just that there's already a builder generator, so probably will have to refactor some things first

metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
jkozlowski: Sure

metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
jkozlowski: Alright we gotta builder!

metric-schema-java/src/integrationInput/java/com/palantir/test/NamespaceTagsMetrics.java
jkozlowski: DOne

metric-schema-java/src/main/java/com/palantir/metric/schema/Custodian.java
```diff
@@ -35,7 +35,7 @@ static String sanitizeName(String input) {
         return escapeIfNecessary(sanitized);
     }
 
-    static String anyToUpperCamel(String input) {
+    public static String anyToUpperCamel(String input) {
```
carterkozak: I don't see any new `Custodian` imports, does this class need to be public?

jkozlowski: Good spot, missed it from ealier iteration.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -92,30 +109,196 @@ static JavaFile generateUtilityClass(
...
+                .map(tagDef -> BuilderStage.builder()
+                        .name(tagDef.getName())
+                        .sanitizedName(Custodian.sanitizeName(tagDef.getName()))
+                        .className(getTagClassName(name, tagDef))
+                        .build())
+                .collect(Collectors.toList());
+        CodeBlock.Builder buildBlock = CodeBlock.builder();
+        buildBlock.add("new $T($L", className, metricRegistryStage.sanitizedName());
+        tagStages.forEach(stage -> buildBlock.add(", $L", stage.sanitizedName()));
+        buildBlock.add(");");
```
carterkozak: Just braindumping, no action required unless you want to: javapoet provides codeblock stream collectors to make this sort of thing flow more naturaly, something along the lines of:

```java
buildBlock.addStatement("new $T($L)", className, Stream.concat...map().collect(CodeBlock....);
```

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -92,30 +109,196 @@ static JavaFile generateUtilityClass(
...
+                                .addStatement(
+                                        "this.$1L = $2T.checkNotNull($1L, $3S)",
+                                        tag.sanitizedName(),
+                                        Preconditions.class,
+                                        tag.name() + " is required")
+                                .addStatement("return this")
+                                .build())
+                        .collect(ImmutableList.toImmutableList()))
+                .build());
+        MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder("builder")
```
carterkozak:
```suggestion
        MethodSpec.Builder methodBuilder = MethodSpec.methodBuilder(ReservedNames.BUILDER_METHOD)
```

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -92,30 +109,196 @@ static JavaFile generateUtilityClass(
...
+                .addModifiers(stagedBuilderSpec.visibility().apply())
```
carterkozak:
```suggestion
                .addModifiers(stagedBuilderSpec.visibility().apply(Modifier.STATIC))
```

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
jkozlowski: Done

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
jkozlowski: 🤦

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
jkozlowski: 👨‍🍳

jkozlowski: Almost there, can't figure out toString

jkozlowski: Done

carterkozak:
Love it, I've tried a snapshout out internally and it simplifies some gnarly code :-] :+1:

jkozlowski:
👍

jkozlowski:
> Love it, I've tried a snapshout out internally and it simplifies some gnarly code :-] 👍

I even know which one you mean! It's long overdue, breached my 3 times rule probably 2 times already

PR #1007, opened 2023-06-26
Generate metric name methods for all metrics

pkoenig10:
Continuation of https://github.com/palantir/metric-schema/pull/293, but exposing this capability for all metrics.

I think the risk of misuse is minimal. Consumers can always bypass metric-schema entirely if they want. But if they want/need to off-road, it's better to at least have the metrics documented in metric-schema.

There are a couple places in our internal server library where this would be useful.

### DIFF
--- a/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
+++ b/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
@@ -517,6 +517,7 @@ final class UtilityGenerator {
         MethodSpec abstractBuildMethod = abstractBuildMethodBuilder.build();
         MethodSpec abstractBuildMetricName = MethodSpec.methodBuilder("buildMetricName")
                 .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
+                .addAnnotation(CheckReturnValue.class)
                 .returns(MetricName.class)
                 .build();
 
--- a/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
+++ b/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
@@ -114,11 +114,11 @@ final class UtilityGenerator {
         metrics.getMetrics().forEach((metricName, definition) -> {
             generateConstants(builder, metricName, definition, visibility);
             if (numArgs(definition) <= 1) {
-                builder.addMethods(generateSimpleMetricFactory(
-                        namespace, metricName, libraryName, metrics, definition, visibility));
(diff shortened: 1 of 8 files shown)


metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -506,13 +519,11 @@ private static void generateMetricFactoryBuilder(
...
                 .build();
-        List<MethodSpec> abstractBuildMethods = isGauge
-                ? ImmutableList.of(abstractBuildMethod, abstractBuildMetricName)
-                : ImmutableList.of(abstractBuildMethod);
 
         outerBuilder.addType(TypeSpec.interfaceBuilder(buildStage(metricName))
                 .addModifiers(visibility.apply())
-                .addMethods(abstractBuildMethods)
+                .addMethod(abstractBuildMethod)
+                .addMethod(abstractBuildMetricName)
```
carterkozak, line 527: Looks like an oversight in the initial implementation, we should add a `@CheckReturnValue` annotation to the `buildMetricName` method.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -564,18 +574,19 @@ private static void generateMetricFactoryBuilder(
                     // TODO(ckozak): Update to use a method which can log a warning and replace existing gauges.
                     // See MetricRegistries.registerWithReplacement.
                     .addStatement(
-                            "$L.$L($L(), $L)",
+                            "$L.$L($N(), $L)",
```
carterkozak, line 578: I think these should all have been `$N` because they're name references, shouldn't make much difference in practice outside of cases like this where we pass a *Spec. Up to you if you'd like to replace them here or not.

carterkozak (approved):
looks good to me, thanks! Feel free to add merge-when-ready at your convenience

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
pkoenig10: Oh interesting, `$N` also [supports](https://github.com/square/javapoet/blob/ba4395e8dae03738a1768d7f6ee02e78fbf44faf/src/main/java/com/squareup/javapoet/CodeBlock.java#L339) `CharSequence`. I assumed it was just for `Spec` types because the [docs](https://github.com/square/javapoet/tree/master#n-for-names) say "Use `$N` to refer to another generated declaration by its name."

pkoenig10:
👍

Issue #1016, opened 2023-07-09
Adding docs to TagValue enumeration values


nickbar01234:
Some enumeration values are not self-explanatory and requires additional documentation to understand metric definitions. Currently, it's not possible to do so because `metrics-schema-api.yml` defines `TagValue` as

```yml
TagValue:
  fields:
    value: string
```

My work-around so far is to add the documentation for enumeration values in `docs` for `TagDefinition`, but this feels a bit clunky. Would it be possible / make-sense to re-define the schema definition to the following and update the markdown generator to output documentations for enumeration values?

```yml
TagValue:
  fields:
    value: string
    docs: optional<Documentation>
```

carterkozak:
That sounds like a great improvement, I've also wanted to document individual tag values! Would you like to contribute this feature?

In addition to the markdown generator, this would be helpful to include as javadoc on our generated tag value enumerations!

nickbar01234:
Yup, I can take a stab at implementing it!

nickbar01234:
Hey @carterkozak, following up on your comment about Javadoc. The current [code generation](https://github.com/palantir/metric-schema/blob/develop/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java#L71) only builds `enum class` when there are at least 2 enumeration values. Otherwise, the tag value is statically defined in the output code, which makes it hard to add Javadoc to a specific part of the output code. Do you think it makes sense to output an `enum class` if there is at least 1 tag value? 

Also, I'm not very familiar with Javadoc, I would love to hear your thoughts on what I'm planning to do 😁. I imagine the JavaDoc for this metric-schema

```yml
metric:
  tags:
    - name: enum
    - values:
         -  FOO
         -  value: BAR
             docs: Some documentation
```

would yield

```java
// FOO doesn't have @see because there's no documentations
/**
 * @see Metric_Enum#BAR
 */
public enum Metric_Enum {
    FOO("foo")
    /**
     * Some documentation
     */
    BAR("bar")
    ...
}
```

nickbar01234:
Resolved in #1021.

PR #1021, opened 2023-07-19
Allow optional documentations for tag enumerations

nickbar01234:
The changes to this commit addresses issue #1016.

## Before this PR

Some tag value enumerations require additional documentations to understand their purposes. The workaround so far is to add documentations for tag values in the `docs` field for `TagDefinition`. While this work, it feels clunky and does not work great when developers want to find out what individual enums represent.

## After this PR

Users can now optionally define documentation for individual tag value enumerations. If a tag value enumerations contain multiple values, individual `TagValue` with `docs` defined field will have their documentation rendered in Markdown and generated Java code. The changes are compatible for existing clients.

==COMMIT_MSG==
Allow optional documentations for tag enumerations
==COMMIT_MSG==

## Are Docs needed?

We decided to remove sorting `TagValue` enumeration in the output Markdown, which will cause diff logs for `metrics.md`.

### DIFF
--- a/metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
+++ b/metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
@@ -121,7 +121,6 @@ public final class MarkdownRenderer {
                 output.append(" values ")
                         .append(tagDefinition.getValues().stream()
                                 .map(TagValue::getValue)
-                                .sorted()
                                 .collect(Collectors.joining("`,`", "(`", "`)")));
             }
 
@@ -131,13 +130,11 @@ public final class MarkdownRenderer {
             output.append("\n");
 
             if (hasEnumValueDocs) {
-                tagDefinition.getValues().stream()
-                        .sorted(Comparator.comparing(TagValue::getValue))
-                        .forEach(value -> {
-                            output.append(String.format("    - `%s`", value.getValue()));
-                            value.getDocs().ifPresent(docs -> output.append(String.format(": %s", docs)));
-                            output.append("\n");
-                        });
+                tagDefinition.getValues().forEach(value -> {
+                    output.append(String.format("    - `%s`", value.getValue()));
+                    value.getDocs().ifPresent(docs -> output.append(String.format(": %s", docs)));
+                    output.append("\n");
+                });
             }
         });
     }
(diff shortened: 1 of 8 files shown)


metric-schema-markdown/src/main/java/com/palantir/metric/schema/markdown/MarkdownRenderer.java
```diff
@@ -119,6 +119,13 @@ private static void renderLine(
                     output.append(": ").append(docs);
                 });
                 output.append("\n");
+
+                tagDefinition.getValues().forEach(value -> {
+                    value.getDocs()
+                            .ifPresent(_doc -> output.append(String.format(
+                                    "    - `%s`: %s\n",
+                                    value.getValue(), value.getDocs().get())));
+                });
```
carterkozak: I think it's a bit odd that we describe the list of tag values without documentation, then list them again with documentation (example from the test below):

> - `result` values (`failure`,`success`)
>  - `success`: This is a success enum

perhaps when any enum value has documentation, we do away with the `values (each, tag, value)` component and list them all in the new list form as either "- `tagValue`: with documentation" or "- `tagValue`" (without documentation) based on the definition.

I don't have a strong opinion whether we should list all tag values in the new list form, vs only doing so when at least one includes documentation. What do you think?

nickbar01234: I'm in favour of rendering per line only when any enum value has documentation to keep it more concise.

carterkozak: Sounds good. That has the additional upside of not producing a diff in markdown when this is released, and only changing the format when humans add documentation :-)

metric-schema-markdown/src/test/java/com/palantir/metric/schema/markdown/MarkdownRendererTest.java
```diff
@@ -227,8 +239,11 @@ void testComplexTagged() {
                         + "namespace docs\n"
                         + "- `namespace.metric` (meter): metric docs\n"
                         + "  - `namespaceTag`\n"
-                        + "  - `result` values (`failure`,`success`)\n"
-                        + "  - `endpoint`: Some docs");
+                        + "  - `result`\n"
+                        + "    - `failure`\n"
+                        + "    - `success`: This is a success enum\n"
+                        + "  - `endpoint`: Some docs\n"
+                        + "  - `processing` values (`bar`,`foo`)");
```
carterkozak: heh I'm not sure why we sorted these originally. I'd support removing the sort call in both code-paths, even though it will produce a bit of a diff as folks upgrade.

nickbar01234: Done :)

carterkozak (approved):
Great work, thank you!

PR #1061, opened 2023-10-23
Cache metric names when possible

pkoenig10:
## Before this PR

Calling the metric or metric name method constructs a new `MetricName` instance on every invocation.

## After this PR

Calling the metric or metric name method will reuse an existing `MetricName` when there are no metric tags.
- If there are no namespace tags, then the cached `MetricName` will be a static field
- If there are namespace tags, then the cache `MetricName` will be a non-static field

I didn't bother handling the case where all metric tags have a single value because it's a relatively obscure case and would add non-trivial complexity to the implementation.

### DIFF
--- a/metric-schema-java/src/integrationInput/java/com/palantir/test/MyNamespaceMetrics.java
+++ b/metric-schema-java/src/integrationInput/java/com/palantir/test/MyNamespaceMetrics.java
@@ -20,6 +20,13 @@ public final class MyNamespaceMetrics {
     private static final String LIBRARY_VERSION =
             Objects.requireNonNullElse(MyNamespaceMetrics.class.getPackage().getImplementationVersion(), "unknown");
 
+    private static final MetricName workerUtilizationMetricName = MetricName.builder()
+            .safeName("com.palantir.very.long.namespace.worker.utilization")
+            .putSafeTags("libraryName", LIBRARY_NAME)
+            .putSafeTags("libraryVersion", LIBRARY_VERSION)
+            .putSafeTags("javaVersion", JAVA_VERSION)
+            .build();
+
     private final TaggedMetricRegistry registry;
 
     private MyNamespaceMetrics(TaggedMetricRegistry registry) {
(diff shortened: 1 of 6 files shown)


metric-schema-java/src/integrationInput/java/com/palantir/test/MyNamespaceMetrics.java
```diff
@@ -20,6 +20,13 @@ public final class MyNamespaceMetrics {
     private static final String LIBRARY_VERSION =
             Objects.requireNonNullElse(MyNamespaceMetrics.class.getPackage().getImplementationVersion(), "unknown");
 
+    private static final MetricName workerUtilizationMetricName = MetricName.builder()
```
pkoenig10, line 23: I'm using camel case for all fields, even those that are static, to keep the implementation simpler. While not the most idiomatic Java code, I imagine it's acceptable here.

carterkozak: Sounds good to me, worth a comment in the codegen along these lines (if there isn’t already, I haven’t read through the change yet, will review tomorrow).

carterkozak:
I wonder if it would be worthwhile to build a base-metric-name of all the known elements (safeName, library name/version tags, even possibly single-value tags) from which the builder could use builder().from(base), or return directly if no additional data is required?

pkoenig10:
That feels like overkill to me. It's harder to write the codegen for that and it results in less readable output.

metric-schema-java/src/integrationInput/java/com/palantir/test/MyNamespaceMetrics.java
schlosna: I'm supportive of this naming as these aren't compile time constants.

schlosna (approved):
This seems reasonable to me and should avoid some allocations on hot paths.

PR #1092, opened 2024-02-07
MetricSchemaPlugin declares sourcesJar dependency on metric generation

schlosna:
## Before this PR

Gradle 8 excavator https://github.com/palantir/tritium/pull/1861 [is failing](https://app.circleci.com/pipelines/github/palantir/tritium/3293/workflows/bbf0278a-b7a2-4a03-a8d4-4efd78e301e5/jobs/20615) due to missing task dependency for `sourcesJar` task on metrics generation.

```
> Task :tritium-lib:sourcesJar FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Some problems were found with the configuration of task ':tritium-lib:sourcesJar' (type 'Jar').
  - Gradle detected a problem with the following location: '/home/circleci/project/tritium-lib/build/metricSchema/generated_src'.
    
    Reason: Task ':tritium-lib:sourcesJar' uses this output of task ':tritium-lib:generateMetrics' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':tritium-lib:generateMetrics' as an input of ':tritium-lib:sourcesJar'.
      2. Declare an explicit dependency on ':tritium-lib:generateMetrics' from ':tritium-lib:sourcesJar' using Task#dependsOn.
      3. Declare an explicit dependency on ':tritium-lib:generateMetrics' from ':tritium-lib:sourcesJar' using Task#mustRunAfter.
    
    For more information, please refer to https://docs.gradle.org/8.5/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
  - Gradle detected a problem with the following location: '/home/circleci/project/tritium-lib/build/metricSchema'.
    
    Reason: Task ':tritium-lib:sourcesJar' uses this output of task ':tritium-lib:compileMetricSchema' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':tritium-lib:compileMetricSchema' as an input of ':tritium-lib:sourcesJar'.
      2. Declare an explicit dependency on ':tritium-lib:compileMetricSchema' from ':tritium-lib:sourcesJar' using Task#dependsOn.
      3. Declare an explicit dependency on ':tritium-lib:compileMetricSchema' from ':tritium-lib:sourcesJar' using Task#mustRunAfter.
    
[...]

PR #1100, opened 2024-02-14
Use task output as Java source directory

pkoenig10:
A correct implementation of what we intended to do in https://github.com/palantir/metric-schema/pull/1092.

Using the task output as the source directory automatically wires up the necessary task dependencies. We don't need every plugin to be aware of the source JAR.


pkoenig10:
I've confirmed that this fixes both the original issue in Tritium and the subsequent issue in internal repos.

schlosna (approved):
Thanks for fixing this @pkoenig10 !

PR #1118, opened 2024-03-11
Project version is provided by gradle

carterkozak:
This avoids looking up implementation package information, which isn't free, but more importantly this preserves original library version info even when code is shaded into another library. Previously, shaded libraries would report the package information from the containing jar, as opposed to the code version used. This applies to custom classloaders as well, not all of which retain implementation-package metadata.

==COMMIT_MSG==
Project version is provided by gradle
==COMMIT_MSG==


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -40,8 +42,16 @@
 
 @CacheableTask
 public abstract class GenerateMetricSchemaTask extends DefaultTask {
-    private final Property<String> libraryName =
-            getProject().getObjects().property(String.class).value(defaultLibraryName());
+    private final Property<String> libraryName = getProject()
+            .getObjects()
+            .property(String.class)
+            .convention(getProject().provider(this::defaultLibraryName));
```
carterkozak, line 51: I think using a convention is more accurate here than setting `value` explicitly. I updated this to evaluate lazily instead of immediately scraping the version when this task is created, because that seemed safer.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -40,8 +42,16 @@
...
+
+    @org.gradle.api.tasks.Optional
+    private final Property<String> libraryVersion = getProject()
+            .getObjects()
+            .property(String.class)
+            .convention(getProject().provider(this::defaultLibraryVersion));
```
carterkozak, line 57: This will require module recompilation any time versions change, but I think the upsides are worth it. What do you think, @CRogers?

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -82,4 +98,10 @@ private String defaultLibraryName() {
         String rootProjectName = getProject().getRootProject().getName();
         return rootProjectName.replaceAll("-root$", "");
     }
+
+    private String defaultLibraryVersion() {
+        String version = Objects.toString(getProject().getRootProject().getVersion());
+        // Gradle returns 'unspecified' when there is no version information
+        return Strings.isNullOrEmpty(version) || "unspecified".equals(version) ? null : version;
+    }
```
carterkozak: We could return null, falling back to the implementationVersion, when the version appears not to be a GA release or RC. That may be easier on caches, but I'm not thrilled with the idea of swapping the version tag implementation based on the presence of a tag. What do you think?

carterkozak: Updated with this change

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
CRogers: afaik the only difference between `convention` and `value` is that if someone later comes and does `property.set(null)`, the convention would be used rather than the property being empty (if value were used). This barely ever (never?) happens so it doesn't really matter - and sometimes you actually want to be able to set the property to empty (depends on usage - probably not in this case).

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/GenerateMetricSchemaTask.java
```diff
@@ -82,4 +101,20 @@ private String defaultLibraryName() {
...
+    @Nullable
+    private String defaultLibraryVersion() {
+        // Gradle returns 'unspecified' when there is no version information, which is not orderable.
+        String version = Objects.toString(getProject().getRootProject().getVersion());
+        return OrderableSlsVersion.safeValueOf(version)
+                // Only provide version data for releases and release candidates, not snapshots. This way we
+                // don't invalidate build caches on each commit.
+                // We prefer passing along version information at this level because it will not be mutated
+                // by shading, where the fallback based on 'package.getImplementationVersion()' will reflect
+                // the shadow jar version.
```
CRogers, line 114: This makes sense. Honestly, I kinda wish we did this at the `version gitVersion()` level as every time you make a commit you need to rebuild the jar, rebuild any dist/containers containing the jar etc even if no code in the jar changed. Big breaking change though.

carterkozak: That would be interesting to try, something like `[most-recent-release][modifier]`, e.g. `1.0.0` -> `1.0.0-snapshot` -> `1.0.0-rc1` -> `1.0.0-rc1-snapshot`/etc.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -82,10 +83,13 @@ static JavaFile generateUtilityClass(
                             Modifier.PRIVATE,
                             Modifier.STATIC,
                             Modifier.FINAL)
-                    .initializer(
-                            "$T.requireNonNullElse($T.class.getPackage().getImplementationVersion(), \"unknown\")",
-                            Objects.class,
-                            className)
+                    .initializer(libraryVersion
+                            .map(version -> CodeBlock.of("$S", version))
```
CRogers, line 88: Stared at this for a while before noticing this bit, with the bit below for back compat.

CRogers:
:+1: LGTM

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
carterkozak: I've added some inline comments to make this clearer :-) (I think I've invalidated your approval though)

CRogers (approved):

PR #1160, opened 2024-03-27
Generate code into conventional location

pkoenig10:
Also refactors the output of the `compileMetricSchema` task so it can be used directly as the resource source directory directly which allows the dependencies to be wired up implicitly, similar to what we did for the Java source directory in https://github.com/palantir/metric-schema/pull/1100.


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CompileMetricSchemaTask.java
```diff
@@ -20,32 +20,43 @@
...
+
+    @Internal
+    public final Provider<RegularFile> getMetricsJsonFile() {
+        return getOutputDir().file(MetricSchemaPlugin.METRICS_JSON_FILE);
+    }
 
     @TaskAction
     public final void action() throws IOException {
-        File output = getOutputFile().getAsFile().get();
-        getProject().mkdir(output.getParent());
```
pkoenig10, line 45: Eliminating the user of `Project` here is also a requirement for supporting the configuration cache.

https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:disallowed_types

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CheckMetricMarkdownTask.java
```diff
@@ -27,50 +27,30 @@
...
+    static final String NAME = "checkMetricsMarkdown";
 
-    @Optional
-    @InputFile
+    @InputFiles
     @PathSensitive(PathSensitivity.RELATIVE)
-    public final Provider<RegularFile> getMarkdownFile() {
-        return ProviderUtils.filterNonExistentFile(getProject(), markdownFile);
-    }
+    public abstract RegularFileProperty getManifestFile();
```
pkoenig10, line 46: Cleaned up most of these task to use managed properties and set the values in the plugin. This is recommended by Gradle.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CheckMetricMarkdownTask.java
```diff
@@ -27,50 +27,30 @@
...
-        return manifestFile;
-    }
+    static final String NAME = "checkMetricsMarkdown";
 
-    @Optional
-    @InputFile
+    @InputFiles
     @PathSensitive(PathSensitivity.RELATIVE)
-    public final Provider<RegularFile> getMarkdownFile() {
-        return ProviderUtils.filterNonExistentFile(getProject(), markdownFile);
```
pkoenig10, line 63: We can replace `@InputFile` with `@InputFiles` so the file isn't required to exist, which allows us to eliminate this `ProviderUtils` workaround.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/CompileMetricSchemaTask.java
```diff
@@ -22,30 +22,38 @@
...
     @OutputFile
-    public abstract RegularFileProperty getOutputFile();
+    public final Provider<RegularFile> getMetricsJsonFile() {
+        return getOutputDir().file(MetricSchemaPlugin.METRICS_JSON_FILE);
+    }
 
     @TaskAction
     public final void action() throws IOException {
-        File output = getOutputFile().getAsFile().get();
-        getProject().mkdir(output.getParent());
```
pkoenig10, line 45: Gradle will automatically create parent directories for `@OutputFile` properties.

From https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/OutputFile.html:

> The parent directory will be created before the task is executed if it does not exist already.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaMarkdownPlugin.java
```diff
@@ -32,33 +34,38 @@ public void apply(Project project) {
...
 
         TaskProvider<CreateMetricsManifestTask> createMetricsManifest =
-                project.getTasks().named(MetricSchemaPlugin.CREATE_METRICS_MANIFEST, CreateMetricsManifestTask.class);
+                project.getTasks().named(CreateMetricsManifestTask.NAME, CreateMetricsManifestTask.class);
 
-        TaskProvider<CheckMetricMarkdownTask> checkMetricsMarkdown = project.getTasks()
-                .register(MetricSchemaPlugin.CHECK_METRICS_MARKDOWN, CheckMetricMarkdownTask.class, task -> {
-                    task.getManifestFile().set(createMetricsManifest.flatMap(CreateMetricsManifestTask::getOutputFile));
-                    task.setMarkdownFile(project.file("metrics.md"));
-                    task.dependsOn(createMetricsManifest);
```
pkoenig10, line 41: This dependency is declared implicitly with the `getManifestFile` property, so this isn't necessary.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaMarkdownPlugin.java
```diff
@@ -32,33 +34,41 @@ public void apply(Project project) {
...
+                    task.getMarkdownFile().set(markdownFile);
                 });
 
-        project.getTasks()
-                .register(MetricSchemaPlugin.GENERATE_METRICS_MARKDOWN, GenerateMetricMarkdownTask.class, task -> {
-                    task.getManifestFile().set(createMetricsManifest.flatMap(CreateMetricsManifestTask::getOutputFile));
-                    task.getOutputFile().set(project.file("metrics.md"));
-                    task.dependsOn(createMetricsManifest);
-                    // Avoid a race when both checkMetricsMarkdown and generateMetricsMarkdown are requested
-                    task.mustRunAfter(checkMetricsMarkdown);
```
pkoenig10, line 50: I've inverted this `mustRunAfter` ordering. If you're running `generate` and `check` you almost certainly want `generate` to run first.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -31,127 +33,90 @@
...
 
     @Override
     public void apply(Project project) {
         project.getPluginManager().apply(JavaLibraryPlugin.class);
-        SourceDirectorySet sourceSet = createSourceSet(project);
+
+        Provider<Directory> generatedJavaDir =
+                project.getLayout().getBuildDirectory().dir("generated/sources/metric-schema/java/main");
+        Provider<Directory> generatedResourcesDir =
+                project.getLayout().getBuildDirectory().dir("generated/sources/metric-schema/resources/main");
```
pkoenig10, line 50: Previously we would generate the `manifest.json` into the `build/metricSchema` directory. But `build/metricSchema` is used as the resource directory, so it's weird/incorrect to add additional files in that directory after `processResources` has run.

It also preventing us from automatically wiring the resources directory up the to `compile` task because we couldn't use the entire directory as the output.

Resources are now put in `generated/sources/metric-schema/resources/main` and `build/metricSchema` is only used for metric-schema internal things like `manifest.json`.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -31,127 +33,90 @@
...
-            }
-        });
+        configureProjectDependencies(project);
     }
 
     private static void configureIdea(
-            Project project, TaskProvider<? extends Task> generateMetricsTask, Provider<Directory> outputDir) {
+            Project project, Provider<Directory> generatedJavaDir, Provider<Directory> generatedResourcesDir) {
         project.getPluginManager().withPlugin("idea", _plugin -> {
-            project.getTasks().named("ideaModule").configure(task -> task.dependsOn(generateMetricsTask));
```
pkoenig10, line 149: We can just remove the IDE task dependencies. We only support building projects using Gradle (which both Eclipse and IntelliJ support) and Gradle has all the proper dependencies.

pkoenig10 (commented):
I've pushed a larger refactor here that cleans up the plugins and tasks here to simplify them, fix some dependency issues, and better follow Gradle best practices.

CRogers (approved):
Thanks for the fixes!

PR #1192, opened 2024-06-10
Generate TaggedMetricRegistry registry() method

schlosna:
## Before this PR

The generated metrics classes do not expose a mechanism to interact with the underlying `TaggedMetricRegistry`. This is needed for example to unregister metrics.

## After this PR
==COMMIT_MSG==
Generate TaggedMetricRegistry registry() method
==COMMIT_MSG==

## Possible downsides?
This breaks encapsulation of the `TaggedMetricRegistry`


carterkozak:
I could go either way on this, given it's uncommon, it may be reasonable to expect code which unregisters metrics to hold a reference to the original registry as well as the generated *Metrics wrapper.

PR #1443, opened 2025-07-24
Properly declare all required dependencies for generated code

aldexis:
## Before this PR
The generated source code was implicitly depending on `com.palantir.safe-logging:safe-logging` and `io.dropwizard.metrics:metrics-core` as could be seen by running `checkImplicitDependencies` from [gradle-baseline](https://github.com/palantir/gradle-baseline/).

## After this PR
==COMMIT_MSG==
Properly declare all required dependencies for generated code
==COMMIT_MSG==

I've added the missing dependencies as well as a test covering that case so everything was properly declared.

Note: This is a similar PR to https://github.com/palantir/gradle-conjure/pull/1652


gradle-metric-schema/src/test/groovy/com/palantir/metric/schema/gradle/MetricSchemaPluginIntegrationSpec.groovy
```diff
@@ -57,22 +59,36 @@ class MetricSchemaPluginIntegrationSpec extends IntegrationSpec {
...
-            }
         }
         """.stripIndent()
+        TestContentHelpers.addVersionsToPropsFile(file('versions.props'), [
+                'com.google.errorprone:error_prone_annotations',
+                'com.palantir.tritium:tritium-registry',
+                'com.palantir.safe-logging:preconditions',
+                'com.palantir.safe-logging:safe-logging',
+                'io.dropwizard.metrics:metrics-core'
+        ])
```
aldexis, line 90: I've also moved to using `TestContentHelpers` rather than forcing explicit versions

bjlaub (approved):

PR #1544, opened 2025-10-23
Do not declare project dependencies if there is no code to generate

aldexis:
## Before this PR
https://github.com/palantir/metric-schema/pull/1443 introduced two missing dependencies. However, because they were declared as `implementation`, this caused some setups to fail the `checkUnusedDependencies` tasks, in the case where there is `metrics.yml` file and thus no code to generate and use the dependencies (this setup exists in services that still generate the `metrics.md` file even if they themselves don't add new metrics)

## After this PR
==COMMIT_MSG==
Do not declare project dependencies if there is no code to generate
==COMMIT_MSG==

It also turns out that `metrics-core` was actually part of the public API of the generated code, so I updated it to be declared as API rather than implementation.

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
@@ -101,7 +101,10 @@ public final class MetricSchemaPlugin implements Plugin<Project> {
 
         configureIdea(project, generatedJavaDir, generatedResourcesDir);
 
-        configureProjectDependencies(project);
+        // Only add dependencies if we're going to generate code
+        if (!metricSchemaSourceDirectorySet.getFiles().isEmpty()) {
+            configureProjectDependencies(project);
+        }
     }
 
     private static void configureIdea(
@@ -125,8 +128,9 @@ public final class MetricSchemaPlugin implements Plugin<Project> {
         project.getDependencies().add("api", "com.palantir.tritium:tritium-registry");
         project.getDependencies().add("api", "com.palantir.safe-logging:preconditions");
         project.getDependencies().add("api", "com.google.errorprone:error_prone_annotations");
+        // Metric types like Gauge are part of the generated code's public API
+        project.getDependencies().add("api", "io.dropwizard.metrics:metrics-core");
         project.getDependencies().add("implementation", "com.palantir.safe-logging:safe-logging");
-        project.getDependencies().add("implementation", "io.dropwizard.metrics:metrics-core");
     }
 
     private String defaultLibraryName(Project project) {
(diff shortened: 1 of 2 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -125,8 +128,9 @@ private static void configureProjectDependencies(Project project) {
         project.getDependencies().add("api", "com.palantir.tritium:tritium-registry");
         project.getDependencies().add("api", "com.palantir.safe-logging:preconditions");
```
aldexis, line 129: Pretty sure this also should be an `implementation` dependency, but I didn't want to change it just here, due to it possibly already being used transitively in downstream projects.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -101,7 +101,10 @@ public void apply(Project project) {
 
         configureIdea(project, generatedJavaDir, generatedResourcesDir);
 
-        configureProjectDependencies(project);
+        // Only add dependencies if we're going to generate code
+        if (!metricSchemaSourceDirectorySet.getFiles().isEmpty()) {
```
aldexis, line 105: I'm unsure whether this is the proper way to do this. Is there a lazy way to do this with providers or something? (I wouldn't think so due to it being for dependency declaration, but checking just in case)

esword: I think because the whole configuration is setup above this should be fine, so long as nothing is dynamically generating sources at a later point?

esword (approved):

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
CRogers: This initially feels not ideal to me, as it's checking where the files exist immediately at configuration time rather than being lazy as possible? It would be bad if this were generated code, perhaps acceptable if this is human authored code (not really sure without diving in metric-schema).

Anyway, the way to do this lazily is with this:

```java
project.getDependencies().addProvider(...)
```

I'd honestly just do this as this is more likely to be correct.

CRogers: Actually, you can't use that `addProvider` to conditionally not a dependency.

Instead you want:

```java
project.getConfigurations().named("api", conf -> {
    conf.getDependencies().addAllLater(project.provider(() -> 
        /* return a list/set of dependencies made with project.getDependencies().create(...) */);
})
```

aldexis: This is checking for manually-authored metrics.yml files in the custom sourceset that is defined above, containing the metrics definitions to generate code from, so this should be mostly ok.

PR #1545, opened 2025-10-23
Lazily configure dependencies

pkoenig10:
Follow up from https://github.com/palantir/metric-schema/pull/1544#discussion_r2454983795.

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
@@ -16,14 +16,17 @@
 
 package com.palantir.metric.schema.gradle;
 
+import com.google.common.collect.ImmutableSetMultimap;
 import com.palantir.sls.versions.OrderableSlsVersion;
 import com.palantir.sls.versions.SlsVersion;
 import com.palantir.sls.versions.SlsVersionType;
+import java.util.List;
 import java.util.Objects;
 import javax.annotation.Nullable;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.file.Directory;
+import org.gradle.api.file.FileCollection;
 import org.gradle.api.file.RegularFile;
 import org.gradle.api.file.SourceDirectorySet;
 import org.gradle.api.plugins.JavaLibraryPlugin;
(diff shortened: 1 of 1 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -99,38 +102,49 @@ public void apply(Project project) {
...
+        configureDependencies(project, metricSchemaSourceDirectorySet);
     }
 
-    private static void configureIdea(
-            Project project, Provider<Directory> generatedJavaDir, Provider<Directory> generatedResourcesDir) {
+    private static void configureIdea(Project project, Provider<Directory> generatedJavaDir) {
         project.getPluginManager().withPlugin("idea", _plugin -> {
             IdeaModel ideaModel = project.getExtensions().getByType(IdeaModel.class);
 
-            ideaModel.getModule().getSourceDirs().add(generatedJavaDir.get().getAsFile());
```
pkoenig10, line 115: This is unnecessary. The IDEA plugin knows the source and resource directories from the source sets. The only thing it can't know is which source directories should be considered generated sources.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
```diff
@@ -99,38 +102,49 @@ public void apply(Project project) {
...
+                                metricSchemaSourceDirectorySet.getElements().map(files -> {
+                                    // Don't add dependencies if we're not going to generate code
+                                    if (files.isEmpty()) {
+                                        return List.of();
+                                    }
+
+                                    return dependencies.stream()
+                                            .map(project.getDependencies()::create)
+                                            .toList();
+                                }));
```
aldexis, line 145: If we end up doing this, I wonder if it would be possible to inspect the outputs from `GenerateMetricsTask` instead, as this is the actual codegen task, which means we could actually check whether files were generated, rather than rely on the inputs existing. Both should work, but the latter seems semantically more correct

pkoenig10: The current implementation seems better to me - I don't think we want dependency resolution to depend on the result of running tasks. If we can know this information without executing the task, that seems preferrable.

aldexis (approved):

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/MetricSchemaPlugin.java
pkoenig10: I tried this in https://github.com/palantir/metric-schema/commit/b8c9b8f562fb6a9bae07afb8000af070a4fec164.

It looks like this doesn't set up a task dependency, so dependency resolution just fails if the task has not yet been run and the output directory does not exist.

https://app.circleci.com/pipelines/github/palantir/metric-schema/5077/workflows/0e731f44-c7f7-4cfb-a789-dbd257cf68c7/jobs/16689/tests

```
org.gradle.api.GradleException: Build aborted because of an internal error.
	at nebula.test.functional.internal.DefaultExecutionResult.rethrowFailure(DefaultExecutionResult.groovy:97)
	at nebula.test.Integration$Trait$Helper.runTasksSuccessfully(Integration.groovy:151)
	at com.palantir.metric.schema.gradle.MetricSchemaPluginIntegrationSpec.setup(MetricSchemaPluginIntegrationSpec.groovy:91)
Caused by: org.gradle.internal.exceptions.LocationAwareException: A problem occurred configuring root project 'generates-java'.
	...
Caused by: org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'generates-java'.
	...
Caused by: java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /home/circleci/project/gradle-metric-schema/build/nebulatest/com.palantir.metric.schema.gradle.MetricSchemaPluginIntegrationSpec/generates-java/build/generated/sources/metricSchema/java/main
	at com.palantir.metric.schema.gradle.MetricSchemaPlugin.lambda$configureDependencies$7(MetricSchemaPlugin.java:149)
	...
Caused by: java.nio.file.NoSuchFileException: /home/circleci/project/gradle-metric-schema/build/nebulatest/com.palantir.metric.schema.gradle.MetricSchemaPluginIntegrationSpec/generates-java/build/generated/sources/metricSchema/java/main
	at org.gradle.internal.classpath.declarations.NioFileInterceptors.intercept_list(NioFileInterceptors.java:274)
	at com.palantir.metric.schema.gradle.MetricSchemaPlugin.lambda$configureDependencies$7(MetricSchemaPlugin.java:144)
	...
```

aldexis: That's too bad, but your earlier point makes sense anyway, so this is fine by me

PR #1553, opened 2025-10-27
Ignore safe-logging dep if unused + replace Strings.CS usage

aldexis:
## Before this PR
It seems like https://github.com/palantir/metric-schema/pull/1544 (and subsequent https://github.com/palantir/metric-schema/pull/1545) were not quite enough, as there are cases where we actually generate code, but do not in fact need the safe-logging dependency (specifically if no tags are used).

Additionally, there seems to be a bug where we fail with e.g.
```
class org.apache.commons.lang3.Strings tried to access method 'boolean org.apache.commons.lang3.CharSequenceUtils.regionMatches(java.lang.CharSequence, boolean, int, java.lang.CharSequence, int, int)' (org.apache.commons.lang3.Strings is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @3d27b703; org.apache.commons.lang3.CharSequenceUtils is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @7bbc7fa0)

* Causal chain is:
	org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':resource-policy-enforcer:generateMetrics'.
	java.lang.IllegalAccessError: class org.apache.commons.lang3.Strings tried to access method 'boolean org.apache.commons.lang3.CharSequenceUtils.regionMatches(java.lang.CharSequence, boolean, int, java.lang.CharSequence, int, int)' (org.apache.commons.lang3.Strings is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @3d27b703; org.apache.commons.lang3.CharSequenceUtils is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader$InstrumentingVisitableURLClassLoader @7bbc7fa0)
```
when calling `Strings.CS.appendIfMissing`. Considering how trivial the logic is, it's simpler to just inline it.

## After this PR

==COMMIT_MSG==
Ignore safe-logging dep if unused + replace Strings.CS usage
==COMMIT_MSG==

### DIFF
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
@@ -24,14 +24,11 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Optional;
+import javax.annotation.Nonnull;
 
 public final class DependencyRequirements {
 
-    public static Multimap<String, String> getDependencies(Collection<File> metricSchemaFiles) {
-        if (metricSchemaFiles == null || metricSchemaFiles.isEmpty()) {
-            return ImmutableSetMultimap.of();
-        }
-
+    public static Multimap<String, String> getDependencies(@Nonnull Collection<File> metricSchemaFiles) {
         Collection<LangMetricSchema> rawSchemas = metricSchemaFiles.stream()
                 .<Optional<LangMetricSchema>>map(schemaFile -> {
                     try {
@@ -44,8 +41,7 @@ public final class DependencyRequirements {
                 .<LangMetricSchema>mapMulti(Optional::ifPresent)
                 .toList();
 
-        if (rawSchemas.isEmpty()
-                || rawSchemas.stream().allMatch(schema -> schema.namespaces().isEmpty())) {
+        if (rawSchemas.stream().allMatch(schema -> schema.namespaces().isEmpty())) {
             // There isn't anything to generate, so no dependencies are required.
             return ImmutableSetMultimap.of();
         }
--- a/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
+++ b/gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.palantir.metric.schema;
+package com.palantir.metric.schema.gradle;
 
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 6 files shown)


gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Collection;
+import java.util.Optional;
+
+public final class DependencyRequirements {
+
+    public static Multimap<String, String> getDependencies(Collection<File> metricSchemaFiles) {
+        if (metricSchemaFiles == null || metricSchemaFiles.isEmpty()) {
+            return ImmutableSetMultimap.of();
+        }
+
```
pkoenig10: `metricSchemaFiles` is never null and the empty case is handled by the code below.

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
```diff
@@ -0,0 +1,76 @@
...
+                    } catch (IOException e) {
+                        // for configuration purposes, ignore files that cannot be parsed as schemas
+                        return Optional.empty();
+                    }
+                })
+                .<LangMetricSchema>mapMulti(Optional::ifPresent)
+                .toList();
+
+        if (rawSchemas.isEmpty()
+                || rawSchemas.stream().allMatch(schema -> schema.namespaces().isEmpty())) {
```
pkoenig10: `allMatch` on an empty collection is true.

```suggestion
        if (rawSchemas.stream().allMatch(schema -> schema.namespaces().isEmpty())) {
```

pkoenig10 (dismissed):

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
aldexis: Good points, thanks! I've updated per your comments - will need another +1 to get it to merge though

gradle-metric-schema/src/main/java/com/palantir/metric/schema/gradle/DependencyRequirements.java
```diff
@@ -0,0 +1,72 @@
...
+import com.palantir.metric.schema.lang.MetricSchemaCompiler;
+import java.io.File;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+
+public final class DependencyRequirements {
+
+    public static Multimap<String, String> getDependencies(@Nonnull Collection<File> metricSchemaFiles) {
```
pkoenig10, line 31: nit: We generally assume values are not null, this feels unnecssary

```suggestion
    public static Multimap<String, String> getDependencies(Collection<File> metricSchemaFiles) {
```

pkoenig10 (approved):

PR #1588, opened 2025-11-14
Add support for custom histogram suppliers

invliD:
## Before this PR
Histograms can only be initialized using the registry's default histogram supplier.

## After this PR
==COMMIT_MSG==
Add support for custom histogram suppliers
==COMMIT_MSG==

## Possible downsides?
🤷🏻‍♂️

### DIFF
--- a/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
+++ b/metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
@@ -17,6 +17,7 @@
 package com.palantir.metric.schema;
 
 import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Histogram;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
@@ -41,6 +42,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import javax.lang.model.element.Modifier;
 
@@ -535,6 +537,7 @@ final class UtilityGenerator {
     }
 
     /** Produce a private staged builder, which implements public interfaces. */
+    @SuppressWarnings("checkstyle:MethodLength")
     private static void generateMetricFactoryBuilder(
             TypeSpec.Builder outerBuilder,
             String namespaceName,
@@ -544,6 +547,10 @@ final class UtilityGenerator {
             MetricNamespace metricNamespace,
             ImplementationVisibility visibility) {
         boolean isGauge = MetricType.GAUGE.equals(definition.getType());
+        boolean isHistogram = MetricType.HISTOGRAM.equals(definition.getType());
+
+        TypeSpec.Builder buildStageInterfaceBuilder =
+                TypeSpec.interfaceBuilder(buildStage(metricName)).addModifiers(visibility.apply());
 
         MethodSpec.Builder abstractBuildMethodBuilder = MethodSpec.methodBuilder("build")
                 .addModifiers(Modifier.PUBLIC, Modifier.ABSTRACT)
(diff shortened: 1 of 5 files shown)


metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -595,6 +610,42 @@ private static void generateMetricFactoryBuilder(
...
+                                        Custodian.sanitizeName(tag.getName()),
+                                        tag.getName() + " is already set")
+                                .addStatement(
+                                        "this.$1L = $2T.checkNotNull($1L, $3S)",
+                                        Custodian.sanitizeName(tag.getName()),
+                                        Preconditions.class,
+                                        tag.getName() + " is required")
+                                .addStatement("return this")
+                                .build())
+                        .collect(ImmutableList.toImmutableList()));
```
invliD, line 647: This entire hunk is just moved from below to simplify adding methods conditionally.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -535,6 +537,7 @@ private static void generateSimpleMetricFactory(
     }
 
     /** Produce a private staged builder, which implements public interfaces. */
+    @SuppressWarnings("checkstyle:MethodLength")
```
pkoenig10, line 540:
```suggestion
    @SuppressWarnings("MethodLength")
```

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -555,18 +562,26 @@ private static void generateMetricFactoryBuilder(
         } else {
             abstractBuildMethodBuilder.addAnnotation(CheckReturnValue.class);
         }
-        MethodSpec abstractBuildMethod = abstractBuildMethodBuilder.build();
-        MethodSpec abstractBuildMetricName = MethodSpec.methodBuilder("buildMetricName")
+        buildStageInterfaceBuilder.addMethod(abstractBuildMethodBuilder.build());
+
+        if (isHistogram) {
```
pkoenig10, line 567: `TaggedMetricRegistry` also provider `Supplier` overloads for `timer`, `meter`, and `counter`. Let's also add overloads for those.

pkoenig10 (commented):
LGTM, although I'd like a +1 from someone else just to confirm that these overloads weren't intentionally omitted.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -555,18 +562,26 @@ private static void generateMetricFactoryBuilder(
...
+            buildStageInterfaceBuilder.addMethod(MethodSpec.methodBuilder("build")
```
schlosna, line 568: pedantic: can we use a more descriptive method name and also generate some JavaDoc?

```suggestion
            buildStageInterfaceBuilder.addMethod(MethodSpec.methodBuilder("buildWithHistograms")
```

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
```diff
@@ -628,46 +679,27 @@ private static void generateMetricFactoryBuilder(
...
-                                        Preconditions.class,
-                                        tag.getName() + " is required")
-                                .addStatement("return this")
-                                .build())
-                        .collect(ImmutableList.toImmutableList()))
-                .addMethod(buildMethod)
-                .addMethod(buildMetricName)
-                .build());
+        if (isHistogram) {
+            builderBuilder.addMethod(MethodSpec.methodBuilder("build")
```
schlosna, line 685:
```suggestion
            builderBuilder.addMethod(MethodSpec.methodBuilder("buildWithHistograms")
```

schlosna (commented):
> LGTM, although I'd like a +1 from someone else just to confirm that these overloads weren't intentionally omitted.

I believe original intent was to have metric-schema provide minimal API surface to "do the right thing" by default. 

I'm supportive adding histogram suppliers as they can be useful, but it might help to require opting-in via metric-schema itself to help document the metrics produced. For example, I've seen a few different places use various sliding windows as well as high dynamic range histograms capturing full lifetime, and the resulting metric ranges vary significantly.

@invliD I am curious what type(s) of reservoir you're looking to provide.

metric-schema-java/src/main/java/com/palantir/metric/schema/UtilityGenerator.java
pkoenig10: Personally, I'd prefer if we named this `build` to be consistent with the `TaggedMetricRegistry` API.

I don't think the "WithHistograms" qualifier adds much value - this is obvious from the method signature.

pkoenig10:
> I believe original intent was to have metric-schema provide minimal API surface to "do the right thing" by default.

I'm pretty sympathetic to this argument.

Given how uncommon this use case is, it doesn't seem unreasonable to say that this isn't supported directly by metric-schema. metric-schema does provide the metric name methods, which make it pretty easy to still leverage metric-schema for the metric name generation and then just use the `TaggedMetricRegistry` APIs directly for registering the metric.

invliD:
I will propose a different away of achieving what I need.

